PR #17, opened 2020-05-13
Feat: Added Matomo siteId for our ACC env

robinpiets:
`environment === "production" ? 29 : 37`

New ACC ID = 37

**please feedback**

### DIFF
--- a/apps/client/src/config.js
+++ b/apps/client/src/config.js
@@ -1,6 +1,8 @@
+export const isProduction = "vergunningcheck.nl" === window.location.hostname;
+
 export const matomo = {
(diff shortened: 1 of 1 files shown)


afjlambert:
I think this will not work because our acceptance env is running in production mode too.

robinpiets:
> I think this will not work because our acceptance env is running in production mode too.

You are right...

robinpiets:
> I think this will not work because our acceptance env is running in production mode too.

What do you think of this solution?

image

afjlambert (approved):
Looks good

svenjens (approved):

afjlambert:
Merge it! 😃

PR #18, opened 2020-05-14
Feature/stadsgezichten

afjlambert:


svenjens (approved):

robinpiets:
@afjlambert @svenjens, it would be nice to include in our way of working that we rebase all commits for a PR. This keeps the repo more organised. Can you do that Andre?

PR #26, opened 2020-05-19
Frontend fixes

robinpiets:
- Fix: updated errors on Location Page
- Feat: Added support for old URL + changed copy

### DIFF
--- a/apps/client/src/config.js
+++ b/apps/client/src/config.js
@@ -117,4 +117,4 @@ export { topics };
 // We need a place for general text as well
 // I know this is not the best place
 // For now I will place it here
-export const requiredFieldText = "Dit veld is verplicht";
+export const requiredFieldText = "Dit veld is verplicht.";
--- a/apps/client/src/config.js
+++ b/apps/client/src/config.js
@@ -73,12 +73,12 @@ const topics = [
     intro: "KozijnenIntroTest",
   },
   {
-    slug: "zonnepanelen-of-warmtecollectoren-plaatsen",
+    slug: "zonnepanelen-of-zonneboiler-plaatsen",
     sttrFile: "zonnepaneel.json",
     text: {
-      heading: "Vergunningcheck zonnepanelen of warmtecollectoren plaatsen",
+      heading: "Vergunningcheck zonnepanelen of zonneboiler plaatsen",
       locationIntro:
-        "Voer het adres in waar u de zonnepanelen of warmtecollectoren wilt gaan plaatsen",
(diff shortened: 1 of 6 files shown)


apps/client/src/config.js
```diff
@@ -100,6 +100,12 @@ const topics = [
   },
 ];
 
+// Support for the old URL ("zonnepanelen-of-warmtecollectoren-plaatsen")
+topics.push({
+  ...topics.find((t) => t.slug === "zonnepanelen-of-zonneboiler-plaatsen"),
+  slug: "zonnepanelen-of-warmtecollectoren-plaatsen",
+});
+
```
afjlambert: Can we instead do a redirect from old slugs to new ones?

afjlambert (changes requested):

apps/client/src/config.js
robinpiets: See my Slack message in #frontend

afjlambert: Maybe something like this?

```
diff --git a/apps/client/src/components/Router.js b/apps/client/src/components/Router.js
index 0d82d14..9025bbf 100644
--- a/apps/client/src/components/Router.js
+++ b/apps/client/src/components/Router.js
@@ -1,7 +1,7 @@
 import React from "react";
-import { BrowserRouter, Switch, Route } from "react-router-dom";
+import { BrowserRouter, Switch, Route, Redirect } from "react-router-dom";
 
-import { routeConfig } from "../routes";
+import { routeConfig, redirectConfig } from "../routes";
 import ScrollToTop from "./ScrollToTop";
 
 const Router = (props) => {
@@ -9,6 +9,9 @@ const Router = (props) => {
     <BrowserRouter>
       <ScrollToTop />
       <Switch>
+        {redirectConfig.map(([from, to]) => (
+          <Redirect {...{ from, to }} />
+        ))}
         {routeConfig
           .filter((route) => route.component)
           .map((route, i) => (
diff --git a/apps/client/src/routes.js b/apps/client/src/routes.js
index 0546501..e9c7e26 100644
--- a/apps/client/src/routes.js
+++ b/apps/client/src/routes.js
@@ -27,6 +27,11 @@ export const geturl = (route, params) => {
   return reverse(route, params);
 };
 
+export const redirectConfig = [
+  ["/a", "/b"],
+  ["/c", "/d"],
+];
+
 export const routeConfig = [
   {
     name: "home",
```

robinpiets: Implemented, thanks mate!

svenjens (approved):

afjlambert (approved):

PR #27, opened 2020-05-19
added a loading indicator

svenjens:


apps/client/src/components/VisualStyles.js
```diff
@@ -1,11 +1,12 @@
...
+import { Card } from "@datapunt/asc-ui";
 
 export const FigCaption = styled.figcaption`
   font-style: italic;
 `;
 
-export const Figure = styled.figure`
-  min-height: 1px; /* IE11 Bug */
+export const StyledCard = styled(Card)`
+  min-height: 100px; /* IE11 Bug */
```
robinpiets: ### This broke the IE11 fix

IE11 screenshot:

image

apps/client/src/components/VisualStyles.js
```diff
@@ -1,11 +1,12 @@
 import styled from "styled-components";
+import { Card } from "@datapunt/asc-ui";
 
 export const FigCaption = styled.figcaption`
   font-style: italic;
 `;
 
-export const Figure = styled.figure`
-  min-height: 1px; /* IE11 Bug */
+export const StyledCard = styled(Card)`
```
robinpiets: You replaced the `figure` element but kept the `figcaption` element. This is bad practice, because the `figcaption` element can only be placed as the first or last child of the `figure` element.

apps/client/src/components/Visual.js
```diff
@@ -1,9 +1,12 @@
...
-export default ({ title, ...rest }) => (
-  <Figure>
-    <Img {...rest} />
-    {title && <FigCaption>{title}</FigCaption>}
-  </Figure>
-);
+export default ({ title, ...rest }) => {
+  const [loading, setLoading] = useState(true);
+  return (
+    <StyledCard isLoading={loading}>
```
robinpiets: I understand you choose `Card` because it has a `isLoading` prop, but I think `Card`  is not suitable in this place because a Card is not an image container. Can you think of another solution?

robinpiets (changes requested):

apps/client/src/components/Visual.js
robinpiets: Update: you can use `perceivedLoading` from `asc-ui` to style the loading

svenjens: Nicee

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,26 @@ export const FigCaption = styled.figcaption`
 export const Figure = styled.figure`
   min-height: 1px; /* IE11 Bug */
   margin: 0 0 24px 0;
+  ${({ isLoading, theme, animateLoading, backgroundImage }) =>
```
robinpiets: Shameless copy/paste from `ImageCardStyle` from asc-ui :-P

Why keep `theme, animateLoading, backgroundImage` if it's never defined and passed to this function?

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,26 @@ export const FigCaption = styled.figcaption`
...
+    isLoading
+      ? css`
+          ${perceivedLoading(animateLoading)({ theme })}
```
robinpiets: Shameless copy/paste from `ImageCardStyle` from asc-ui :-P

Remove `animateLoading` and `theme`

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,26 @@ export const FigCaption = styled.figcaption`
...
+            content: "";
+            display: block;
+            width: calc(100% - 30px);
+            height: 30px;
+            ${perceivedLoading(animateLoading)({ theme })}
+          }
+        `
+      : css`
+          background-image: url(${backgroundImage});
+        `}
```
robinpiets: Shameless copy/paste from `ImageCardStyle` from asc-ui :-P

We don't have a background-image

robinpiets (changes requested):
Prepare your PRs better please

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,23 @@ export const FigCaption = styled.figcaption`
...
   margin: 0 0 24px 0;
+  ${({ isLoading, theme, animateLoading }) =>
+    isLoading &&
+    css`
+      ${perceivedLoading(animateLoading)}
+      height: 100px;
+      width: 100%;
+      & > * {
+        display: none;
+      }
```
afjlambert: This piece also hides the FigCaption while the image is loading. I don't think this is what we want. Can you fix that?

afjlambert (changes requested):
Schermafbeelding 2020-05-27 om 12 40 45
Schermafbeelding 2020-05-27 om 12 40 54

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,23 @@ export const FigCaption = styled.figcaption`
 export const Figure = styled.figure`
   min-height: 1px; /* IE11 Bug */
   margin: 0 0 24px 0;
+  ${({ isLoading, theme, animateLoading }) =>
```
robinpiets: We don't use `theme` and `animateLoading` so this can be removed

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,23 @@ export const FigCaption = styled.figcaption`
...
+    isLoading &&
+    css`
+      ${perceivedLoading(animateLoading)}
```
robinpiets: We don't use `theme` and `animateLoading` so this can be removed

apps/client/src/components/VisualStyles.js
```diff
@@ -7,6 +8,23 @@ export const FigCaption = styled.figcaption`
...
+      width: 100%;
+      & > * {
+        display: none;
+      }
+      &::before {
+        content: "";
+        display: block;
+        width: calc(100% - 30px);
+        height: 30px;
+        ${perceivedLoading(animateLoading)({ theme })}
```
robinpiets: We don't use `theme` and `animateLoading` so this can be removed

robinpiets (changes requested):

afjlambert (approved):

PR #28, opened 2020-05-20
Fix: Updated Alert components

robinpiets:
- Replaced Alert with new 'warning' level
- Fixed IE11 bugs (see trello card 498)
- Create Alert wrapper with IE11 fix 
- Updated all `Atoms/` imports

**please merge**

### DIFF
--- a/apps/client/src/components/Error.js
+++ b/apps/client/src/components/Error.js
@@ -1,20 +1,23 @@
 import React from "react";
 import { Alert } from "@datapunt/asc-ui";
+import { ComponentWrapper } from "./Atoms";
 
 const Error = ({ heading, content, stack, children }) => (
-  <Alert
-    level="error"
-    heading={heading || `Er is een fout opgetreden.`}
-    content={content}
-    style={{ marginBottom: 25 }}
-  >
-    {children}
+  <ComponentWrapper>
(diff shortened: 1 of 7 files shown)


apps/client/src/components/Error.js
```diff
@@ -1,20 +1,23 @@
...
-    content={content}
-    style={{ marginBottom: 25 }}
-  >
-    {children}
+  <ComponentWrapper>
+    <Alert
+      level="error"
+      heading={heading || `Er is een fout opgetreden.`}
+      content={content}
+      style={{ display: "block" }} // IE11 Fix
```
afjlambert: It's easy to forget this IE11 fix everywhere we use Alert. Maybe we can have our own small wrapper around Alert in our atoms/components folder until (if ever) this bug is fixed in ASC.

afjlambert (approved):

apps/client/src/components/Error.js
robinpiets: I was thinking not to do this, because sometimes you might need a `flexbox` Alert. But I guess it's better not needing to copy/paste all IE11 fixes for every new Alert.

FIxed

svenjens (approved):

PR #53, opened 2020-05-26
Move sttr files to environment specific folders

afjlambert:
This is work in advance of autofill.

Benefits:
- For easy debugging / development
- No change of accidentally pushing staging sttr configs to production
- Got rid of yargs, security issue
- No more staging sttr-config in repo (outcomes.json)
- `git diff` doesn't complain as much when you are running with non-prod sttr config

Hope you like it 😃


apps/client/scripts/sttr_build/index.js
```diff
@@ -62,10 +57,10 @@ function checkStatus(res) {
 }
 
 (async () => {
-  const permitIds = await fetch(listUrl, { headers })
+  fetch(listUrl, { headers })
```
robinpiets, line 60: Do you need `async` when you remove `await`?

apps/client/scripts/sttr_build/index.js
```diff
@@ -1,29 +1,22 @@
...
 const MAX_PARALLEL = 6;
 const env = process.env.STTR_ENV === "production" ? "PROD" : "STAGING";
+const outputDir = path.join(
+  __dirname,
+  "..",
+  "..",
+  "public",
+  "sttr",
+  env.toLowerCase()
+);
```
robinpiets, line 19: Nice!

apps/client/.gitignore
```diff
@@ -1,5 +1,7 @@
-public/sttr/*.xml
-public/sttr/*.source.json
+public/sttr/staging
```
robinpiets, line 1: So we're never commiting the staging files.. makes sense

apps/client/scripts/sttr_build/index.js
afjlambert: Yes there is an await few lines below.

robinpiets (approved):

robinpiets:
ready to merge!

PR #84, opened 2020-05-29
Flatten repo-tree refactor

afjlambert:


afjlambert:
Please use "squash and merge".

apps/client/src/components/Markdown/renderers/ListRenderer.js
```diff
@@ -1,4 +1,4 @@
 import React from "react";
-import { List } from "../../Atoms";
+import { List } from "../../../atoms";
```
robinpiets, line 2: Would be nice to be able to include from `/src` instead of going back up the tree. So `"atoms"` instead of `"../../../atoms"` Can you fix this in this PR or shall we make this an issue or a card in Trello?

apps/client/src/routes.js
```diff
@@ -90,3 +90,8 @@ export const redirectConfig = [
 export const routes = Object.fromEntries(
   routeConfig.map(({ name, path }) => [name, path])
 );
+
+export const autofillRoutes = {
+  address: routes.location,
+  // map: ...
+};
```
robinpiets, line 97: Do we need this in this PR?

robinpiets (commented):
This PR is going to influence the PR from `browser-support`, I would prefer to wait until that PR is merged. What do you think @svenjens?

apps/client/src/routes.js
afjlambert: Already gone. Forked from the wrong branch, sorry.
I will squash it before merging.

PR #101, opened 2020-06-02
Added browser refresh button support

svenjens:
## Please Squash and Merge instead of normal Merge

image


apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
...
+
+  return (
+    <Context.Provider value={{ ...data, setData }}>
+      {props.children}
+    </Context.Provider>
+  );
+}
+
+export default Context;
+export { defaultValues, CheckProvider };
```
afjlambert: `defaultValues` is no longer used outside of this file right?
Please remove the export if we can.

apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
-import React from "react";
-export const defaultValues = { topic: null, address: null, checker: null };
-export default React.createContext();
+import React, { useReducer, useEffect, createContext } from "react";
```
afjlambert, line 1: Can you add some comments in this file please? For me it's not really clear what we are actually doing here.

apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
...
+    return defaultValues;
+  }
+  return { ...data, ...newData };
+};
+
+const defaultValues = session ? session : { topic: null, address: null };
+
+const Context = createContext();
+
+function CheckProvider(props) {
```
afjlambert: Can we give a more descriptive name, maybe SessionProvider?

apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
...
+
+let reducer = (data, newData) => {
+  if (newData === null) {
+    sessionStorage.removeItem("vergunningCheckData");
+    return defaultValues;
+  }
+  return { ...data, ...newData };
+};
+
+const defaultValues = session ? session : { topic: null, address: null };
```
afjlambert: Can we move this up to the top? It's used in the `reducer` function.

apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
-import React from "react";
-export const defaultValues = { topic: null, address: null, checker: null };
-export default React.createContext();
+import React, { useReducer, useEffect, createContext } from "react";
+const session = JSON.parse(sessionStorage.getItem("vergunningCheckData"));
+
+let reducer = (data, newData) => {
```
afjlambert: Can this be a const?

apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
-import React from "react";
-export const defaultValues = { topic: null, address: null, checker: null };
-export default React.createContext();
+import React, { useReducer, useEffect, createContext } from "react";
+const session = JSON.parse(sessionStorage.getItem("vergunningCheckData"));
```
afjlambert: Can we use an English variable name here? Also we never use camelCase for "vergunningcheck" as far as I know.

apps/client/src/context.js
```diff
@@ -1,3 +1,31 @@
...
+
+function CheckProvider(props) {
+  const [data, setData] = useReducer(reducer, defaultValues);
+
+  useEffect(() => {
+    sessionStorage.setItem("vergunningCheckData", JSON.stringify(data));
+  }, [data]);
+
+  return (
+    <Context.Provider value={{ ...data, setData }}>
```
afjlambert: Good place to add a comment here?

apps/client/src/pages/LocationPage.js
```diff
@@ -40,7 +40,10 @@ const LocationPage = ({ topic }) => {
         action: `postcode - ${slug.replace("-", " ")}`,
         name: address.postalCode.substring(0, 4),
       });
-      context.address = address;
+      context.setData({
+        address,
+        data: null,
```
afjlambert: I'm confused. In `defaultValues` there is `{ topic: null, address: null }`. There is no `data` prop.
Here we have a `setData(data)` where `data` has a `data` property.

apps/client/src/sttr_client/models/checker.js
```diff
@@ -50,6 +50,33 @@ class Checker {
     return this._stack;
   }
 
+  /**
+   * Get data from the stack and return a list of questionIds with the given answers.
+   * @returns {Answers[]} -
```
afjlambert: This function returns an object with "question.id" as keys and "answer" values.
So `string: (boolean|string|number|string[])` I guess.

apps/client/src/pages/QuestionsPage.js
```diff
@@ -47,6 +49,9 @@ const QuestionsPage = ({ topic, checker }) => {
 
   const onQuestionNext = (value) => {
     if (question.options) {
+      context.setData({
```
afjlambert: Can you add a comment why you need to set the data on context before you set the answer to the question. Is it because of position on the stack?

apps/client/src/context.js
afjlambert, line 1: Maybe context.js is a good candidate to convert to context.ts?

afjlambert (changes requested):
Looks good @svenjens.
Couple of questions though.

apps/client/src/sttr_client/models/checker.js
```diff
@@ -50,6 +50,33 @@ class Checker {
...
+   */
+  getData() {
```
robinpiets: Shall we rename this to `getQuestionAnswers()`? @svenjens @afjlambert 

This is too general and we're probably going to need more of these getters/setters in the near future

apps/client/src/sttr_client/models/checker.test.js
```diff
@@ -17,28 +17,32 @@ const q2 = new Question({
...
+    [q1, q2],
+    [new Rule([true, false], "non local"), new Rule([true, true], "local")]
+  );
+  const d3 = new Decision(
+    "dummy",
+    [d1, d2],
+    [
+      new Rule(["boring"], "Maybe you should move?"),
+      new Rule(["fun!", "non local"], "Hi Robin or Sven"),
+      new Rule(["fun!", "local"], "Hi André"),
```
robinpiets, line 37: I would prefer it if we change these rules to something more general and understanding, like `Answer option 1` and `Question title 1`

apps/client/src/pages/QuestionsPage.js
```diff
@@ -7,9 +7,11 @@ import withChecker from "../hoc/withChecker";
 import Layout from "../components/Layouts/DefaultLayout";
 import DebugDecisionTable from "../components/DebugDecisionTable";
 import Question, { booleanOptions } from "../components/Question";
+import Context from "../context";
 // import ErrorPage from "./ErrorPage";
```
robinpiets: Can you start with the reordering of imports? And also remove the unused import?

apps/client/src/context.js
```diff
@@ -1,3 +1,34 @@
...
+function SessionProvider(props) {
+  // We use the reducer to take care of too complex logic for setState.
+  // Because we sometimes need to clear the sessionStore.
+  const [data, setData] = useReducer(reducer, defaultValues);
+
+  useEffect(() => {
+    sessionStorage.setItem("sessionData", JSON.stringify(data));
+  }, [data]);
+
+  // The session provider makes the data from the context avaible on all pages.
```
robinpiets: typo: avaible

apps/client/src/context.js
```diff
@@ -1,3 +1,34 @@
...
+  if (newData === null) {
+    sessionStorage.removeItem("sessionData");
+    return defaultValues;
+  }
+  return { ...data, ...newData };
+};
+
+function SessionProvider(props) {
+  // We use the reducer to take care of too complex logic for setState.
+  // Because we sometimes need to clear the sessionStore.
```
robinpiets: np: sessionStore > sessionStorage

apps/client/src/hoc/withAddress.js
```diff
@@ -8,11 +8,12 @@ const withAddress = (Component) =>
   withTopic(({ ...rest }) => {
     const { address } = useContext(Context);
     const { topic } = rest;
+
     if (!address) {
       console.warn("No address found, redirecting to location page");
       return <Redirect to={geturl(routes.location, { slug: topic.slug })} />;
     }
-    return <Component address={address} {...rest} />;
+    return <Component address={address?.[topic.slug]} {...rest}></Component>;
```
robinpiets: Cool!!!

apps/client/src/pages/QuestionsPage.js
```diff
@@ -7,20 +7,24 @@ import withChecker from "../hoc/withChecker";
...
 
 const QuestionsPage = ({ topic, checker }) => {
+  const context = useContext(Context);
   const params = useParams();
   const history = useHistory();
+  const { question: questionSlug } = params;
   const [question, setQuestion] = useState(
-    checker.stack[checker.stack.length - 1]
+    checker.stack[context.questionIndex]
```
robinpiets: This is also a good place for a comment

apps/client/src/pages/ResultsPage.js
```diff
@@ -11,12 +11,15 @@ import Nav from "../components/Nav";
 import QuestionAnswerTable from "../components/QuestionAnswerTable";
 import DebugDecisionTable from "../components/DebugDecisionTable";
 import { Helmet } from "react-helmet";
```
robinpiets: Can we move the `Helmet` import up?

apps/client/src/pages/ResultsPage.js
```diff
@@ -11,12 +11,15 @@ import Nav from "../components/Nav";
...
+import Context from "../context";
 
 const ResultsPage = ({ topic, checker }) => {
   const history = useHistory();
+  const context = useContext(Context);
   const { slug } = topic;
 
   const onGoToQuestion = (index) => {
+    context.setData({ questionIndex: index});
```
robinpiets: Better to rename `index` to `questionIndex`

apps/client/src/sttr_client/models/checker.js
```diff
@@ -50,6 +50,33 @@ class Checker {
...
+      acc[question.id] = question.answer;
+      return acc;
+    }, {});
+  }
+
+  /**
+   * Set data from context to the current stack
+   * @returns {void} -
+   */
+  setData(answers) {
```
robinpiets: Shall we rename this to `setQuestionAnswers()`? @svenjens @afjlambert 

This is too general and we're probably going to need more of these getters/setters in the near future

apps/client/src/utils/test-utils.js
```diff
@@ -4,23 +4,22 @@ import { render } from "@testing-library/react";
 import { ApolloProvider } from "@apollo/react-hooks";
 import { MatomoProvider, createInstance } from "@datapunt/matomo-tracker-react";
 import { ThemeProvider } from "@datapunt/asc-ui";
-
-import Context, { defaultValues } from "../context";
+import { SessionProvider } from "../context";
```
robinpiets, line 7: np: add empty line between external and internal imports

robinpiets (changes requested):

apps/client/src/pages/QuestionsPage.js
svenjens: i agree

apps/client/src/hoc/withChecker.js
```diff
@@ -1,29 +1,38 @@
...
       // TODO: doesn't withAddress already guarantee this? :-/
       console.warn("Address not found, redirecting to location page");
       return <Redirect to={geturl(routes.location, { slug: topic.slug })} />;
     }
 
     useEffect(() => {
+      // If `sttr-checker` has already been loaded but the answers have been cleared
+      if (checker && !sessionContext.answers) {
+        console.log("Reset the checker");
+        // @TODO: How to reset the checker and/or remove the user answers from the stack
```
robinpiets: @afjlambert How to do reset the checker?

apps/client/src/sttr_client/models/checker.js
```diff
@@ -50,6 +50,61 @@ class Checker {
...
+      // The checker is completed when `questionAnswer` is undefined
+      if (questionAnswer === undefined) {
+        done = true;
+      } else {
+        // Answer the question and proceed
+        this._last.setAnswer(questionAnswer);
+        this.next();
+      }
+
+      // Prevent infinite loop, otherwise the browser will explode
```
svenjens: 😂

apps/client/src/sttr_client/models/checker.test.js
```diff
@@ -61,4 +65,14 @@ describe("Checker recursive", () => {
       "Maybe you should move?"
     );
   });
+  test("initialization after refresh", () => {
```
svenjens, line 68: 🙌

apps/client/src/pages/LocationPage.js
```diff
@@ -67,9 +79,9 @@ const LocationPage = ({ topic }) => {
         <LocationFinder
           setAddress={setAddress}
           setErrorMessage={setErrorMessage}
-          postalCode={context.address?.postalCode}
-          houseNumberFull={context.address?.houseNumberFull}
-          houseNumber={context.address?.houseNumberFull}
+          postalCode={context.address?.[slug]?.postalCode}
```
svenjens: I like to see that the data is from the context. 
But i think to best practice is too destructure everything. What do you think?

robinpiets: Fixed. Can you review?

robinpiets (approved):

apps/client/src/context.js
afjlambert: I though you already started working on this. Or not?
Converting this to typescript would also 'fix' the `as any` in consumers of the Contexts.

apps/client/src/components/QuestionAnswerTable.js
```diff
@@ -41,53 +41,59 @@ export default ({ checker, onGoToQuestion }) => {
       </MainWrapper>
 
       <ComponentWrapper marginBottom={40}>
-        {checker?.stack?.map((question, index) => {
+        {checker?.stack?.map((question, questionIndex) => {
```
afjlambert: Why put question marks everywhere? Checker is guaranteed to be here right? Stack is always available on checker.

apps/client/src/components/QuestionAnswerTable.js
```diff
@@ -41,53 +41,59 @@ export default ({ checker, onGoToQuestion }) => {
...
-                  </UserResultParagraph>
-                </UserResult>
-              ))}
-            </QuestionWrapper>
-          );
+            uniqBy(permitsPerQuestion[questionIndex], "name") || [];
+          if (
+            typeof question.answer === "boolean" ||
+            typeof question.answer === "string"
+          ) {
```
afjlambert: Why filter only these questions? What are you not interested in? Consider implementing an `else` case. Even if it's only a warning/notise that `type x` is not supported.

apps/client/src/components/QuestionAnswerTable.js
```diff
@@ -41,53 +41,59 @@ export default ({ checker, onGoToQuestion }) => {
...
+                    </Button>
+                  </Change>
+                )}
+                {isDecisiveForPermits.map((permit, index) => (
+                  <UserResult key={`${permit} ${index}`}>
+                    <Icon
+                      color={themeColor("secondary", "main")}
+                      size={30}
+                      style={{
+                        flexShrink: 0, // IE11 Fix
```
afjlambert: We should consider adding an atom for Icon if it's not safe to use without browser-hacks.
It's too easy to forget to implement such a 'fix'.
Not implemented in this change, I know.

apps/client/src/components/QuestionAnswerTable.js
```diff
@@ -41,53 +41,59 @@ export default ({ checker, onGoToQuestion }) => {
...
+                  <Change>
+                    <Button
+                      onClick={() => onGoToQuestion(questionIndex)}
+                      variant="textButton"
+                    >
+                      Wijzig
+                    </Button>
+                  </Change>
+                )}
+                {isDecisiveForPermits.map((permit, index) => (
```
afjlambert: isDecisiveForPermits sounds like a bool but it's actually an array :-/
Not implemented in this change, I know.

apps/client/src/hoc/withAddress.js
afjlambert: Why the question-mark? Can't we either guarantee it's an Object by declaring it above like `const { address } = useContext(SessionContext) || {};` or make sure it's added in the default session context?

apps/client/src/hoc/withChecker.js
```diff
@@ -1,29 +1,32 @@
...
-    const context = useContext(Context);
-    const [checker, setChecker] = useState(context.checker);
+    const sessionContext = useContext(SessionContext);
+    const checkerContext = useContext(CheckerContext);
+    const [checker, setChecker] = useState(checkerContext.checker);
     const [error, setError] = useState();
 
     const { topic } = rest;
-    if (!context.address) {
+    if (!sessionContext.address) {
```
afjlambert: But sessionContext.address is actually an Object right? Shouldn't you check for `[topic.slug]` as well?

apps/client/src/index.js
```diff
@@ -8,13 +8,13 @@ import { createGlobalStyle } from "styled-components";
 import dotenv from "dotenv-flow";
 import { MatomoProvider, createInstance } from "@datapunt/matomo-tracker-react";
 import { GlobalStyle, ThemeProvider, themeColor } from "@datapunt/asc-ui";
+import "@datapunt/asc-assets/static/fonts/fonts.css";
```
afjlambert: Huh? What does this have to do with refresh-button?
This has never been here? Why add it now?

apps/client/src/pages/ConclusionPage.js
```diff
@@ -34,7 +34,7 @@ const ConclusionPage = ({ topic, checker }) => {
     houseNumberFull,
     postalCode,
     residence,
-  } = context.address;
+  } = sessionContext.address;
```
afjlambert: Again, not sure why/when address is an map with an address per topic vs just the address itself.

apps/client/src/pages/LocationPage.js
```diff
@@ -17,12 +17,14 @@ import Error from "../components/Error";
...
   const { trackEvent } = useMatomo();
-  const context = useContext(Context);
+  const sessionContext = useContext(SessionContext);
+  const checkerContext = useContext(CheckerContext);
   const history = useHistory();
   const [address, setAddress] = useState(null);
   const [errorMessage, setErrorMessage] = useState();
   const { clearError, errors, register, unregister, handleSubmit } = useForm();
   const { slug, text } = topic;
+  const sessionAddress = sessionContext.address?.[slug];
```
afjlambert: Again, not sure. Also question mark again.

apps/client/src/pages/LocationPage.js
```diff
@@ -40,7 +42,21 @@ const LocationPage = ({ topic }) => {
...
+      if (answers && sessionAddress?.id !== address?.id) {
+        checkerContext.checker = null;
+        answers = null;
+      }
+
+      sessionContext.setSessionData({
+        address: { ...sessionContext.address, [slug]: address },
+        answers, // Either null or filled with given answers
+        questionIndex: 0, // Reset to 0 to start with the first question
+      });
```
afjlambert: I think we should consider if this logic should be part of the LocationPage.

apps/client/src/pages/LocationPage.js
```diff
@@ -67,14 +83,16 @@ const LocationPage = ({ topic }) => {
         <LocationFinder
           setAddress={setAddress}
           setErrorMessage={setErrorMessage}
-          postalCode={context.address?.postalCode}
-          houseNumberFull={context.address?.houseNumberFull}
-          houseNumber={context.address?.houseNumberFull}
+          postalCode={sessionAddress?.postalCode}
```
afjlambert: Can we make sure sessionAddress is an `{}`?

apps/client/src/pages/QuestionsPage.js
```diff
@@ -1,40 +1,44 @@
...
+  const { slug } = topic;
 
-  const { question: questionSlug } = params;
-  const currSlug = getslug(question.text);
+  useEffect(() => {
+    if (!questionSlug) {
+      sessionContext.setSessionData({
+        answers: checker.getQuestionAnswers(),
+      });
+    }
```
afjlambert: Can you add a comment?

apps/client/src/pages/QuestionsPage.js
```diff
@@ -1,40 +1,44 @@
...
+  });
 
   // Update URL based on question text
-  if (!questionSlug || questionSlug !== currSlug) {
+  if (!questionSlug) {
```
afjlambert: why remove the logic to redirect if the url-slug does not correspond with the question-slug?

apps/client/src/pages/QuestionsPage.js
```diff
@@ -46,34 +50,78 @@ const QuestionsPage = ({ topic, checker }) => {
     });
 
   const onQuestionNext = (value) => {
-    if (question.options) {
+    // Provide the user answers to the `sttr-checker`
+    if (question.options && value) {
```
afjlambert: Can you be a bit more explicit and check answers if the are not `undefined`?

apps/client/src/pages/QuestionsPage.js
```diff
@@ -46,34 +50,78 @@ const QuestionsPage = ({ topic, checker }) => {
...
       question.setAnswer(value);
-    } else {
+    }
+    if (!question.options && value) {
```
afjlambert, line 49: How can this even work? Value can be `false` right?

apps/client/src/pages/QuestionsPage.js
```diff
@@ -46,34 +50,78 @@ const QuestionsPage = ({ topic, checker }) => {
...
       const responseObj = booleanOptions.find((o) => o.formValue === value);
       question.setAnswer(responseObj.value);
     }
 
-    if (needContactPermits()) {
+    // Store all answers in the session context
```
afjlambert, line 54: Nice optimization would be to implement the syncronization logic inside the checker itself.
Could be an onSetAnser callback. Or give it a 'store' to put the data in and read from.

apps/client/src/pages/QuestionsPage.js
```diff
@@ -46,34 +50,78 @@ const QuestionsPage = ({ topic, checker }) => {
...
+      } else {
+        // Go to Result page if there is no new quesion
+        history.push(geturl(routes.results, { slug }));
       }
     }
   };
 
   const onQuestionPrev = () => {
+    // Load the previous question or go to the Location Page
     if (checker?.stack?.length > 1) {
```
afjlambert: Checker and checker.stack are guaranteed to be here. Please remove the ?'s.

apps/client/src/pages/ResultsPage.js
```diff
@@ -10,14 +12,20 @@ import Form from "../components/Form";
...
+  const sessionContext = useContext(SessionContext);
   const { slug } = topic;
 
-  const onGoToQuestion = (index) => {
-    const q = checker.rewindTo(index);
+  const onGoToQuestion = (questionIndex) => {
+    // Go to the specific question in the stack
+    const q = checker.rewindTo(questionIndex);
+    sessionContext.setSessionData({
+      questionIndex: questionIndex,
```
afjlambert: you can remove the `: questionIndex`.
Doesn't our linter fix this? :-/

apps/client/src/sttr_client/models/checker.js
```diff
@@ -50,6 +50,66 @@ class Checker {
...
+      .reduce((acc, question) => {
+        acc[question.id] = question.answer;
+        return acc;
+      }, {});
+  }
+
+  /**
+   * Loops through all questions and answers these questions based on provided `answers`
+   * Useful when you have stored the user answers and want to continue a particular session
+   * @returns {void} -
```
afjlambert: remove the @returns here please.

apps/client/src/sttr_client/models/checker.js
```diff
@@ -50,6 +50,66 @@ class Checker {
...
+      counter++;
+      if (counter > 1000) {
+        console.error(
+          "Infinite loop detected and prevented in setQuestionAnswers()"
+        );
+        done = true;
+      }
+    }
+  }
+
```
afjlambert: What?

afjlambert (changes requested):

apps/client/src/context.js
svenjens: Let's decide this after the merge with autofill. We might need to refactor some more.

apps/client/src/context.js
```diff
@@ -1,3 +1,47 @@
...
+  ? session
+  : {
+      address: {},
+      answers: null,
+      questionIndex: 0,
+    };
+
+const reducer = (data, newData) => {
+  if (newData === null) {
+    // Clear the current sessionData if incoming data is empty (instead of setting an empty item).
```
svenjens, line 20: 🙌

apps/client/src/hoc/withAddress.js
```diff
@@ -1,18 +1,18 @@
...
+import { SessionContext } from "../context";
 import { routes, geturl } from "../routes";
 import withTopic from "./withTopic";
 
 const withAddress = (Component) =>
   withTopic(({ ...rest }) => {
-    const { address } = useContext(Context);
+    const { address } = useContext(SessionContext);
     const { topic } = rest;
     if (!address) {
```
svenjens: Maybe we should check for the topic slug in address here.

apps/client/src/hoc/withChecker.js
svenjens: We shouldn't need this check right? It's in the withAddress let's check it there.

apps/client/src/hoc/withTopic.js
```diff
@@ -1,27 +1,27 @@
...
   const { search } = useLocation();
 
   const topic = topics.find((t) => t.slug === slug);
   const params = new URLSearchParams(search);
 
   if (params.get("resetChecker")) {
-    context.checker = null;
+    checkerContext.checker = null;
     // TODO: Remove this warning?
     console.warn("Resseting checker, redirecting to intro page");
```
svenjens, line 19: Also, what is resseting.

apps/client/src/index.js
svenjens: it's moved from the bottom to have our import order. 
But i think it should be set on top.

apps/client/src/pages/ResultsPage.js
```diff
@@ -10,14 +12,20 @@ import Form from "../components/Form";
...
 const ResultsPage = ({ topic, checker }) => {
   const history = useHistory();
+  const sessionContext = useContext(SessionContext);
   const { slug } = topic;
 
-  const onGoToQuestion = (index) => {
-    const q = checker.rewindTo(index);
+  const onGoToQuestion = (questionIndex) => {
+    // Go to the specific question in the stack
+    const q = checker.rewindTo(questionIndex);
```
svenjens: let's call this question instead of q

apps/client/src/context.js
robinpiets: Good idea, I think it's better to do this after out merge with `autofill`

apps/client/src/components/QuestionAnswerTable.js
robinpiets: I don't know why this is. @svenjens why did you add this and can we not remove it? The `sttr-build` is providing us with this information right. For now i've added:
```
} else {
  console.warn(
    `The type "${typeof question.answer}" of "${
      question.answer
    }" is not supported`
  );
}
```

robinpiets: Update: removed this check completely

apps/client/src/components/QuestionAnswerTable.js
robinpiets: renamed to `decisivePermits` @afjlambert @svenjens

apps/client/src/hoc/withChecker.js
robinpiets: @afjlambert `sessionContext.address` is either `null` or an object with data. If `sessionContext.address` exists it means that the address has been set. But as @svenjens  says, it's already been placed in the `withAddress`

apps/client/src/index.js
robinpiets: Indeed, it was at the bottom. I will put it back and will leave it up to the eslint / prettier to deal with import order. For now resolving this

apps/client/src/hoc/withAddress.js
robinpiets: It's gone now

apps/client/src/pages/ConclusionPage.js
robinpiets: This was actually a bug. It needed `address[slug]`

apps/client/src/pages/LocationPage.js
robinpiets: This questionmark needs to be there otherwise you get a `TypeError: Cannot read property 'dakkapel-plaatsen' of null` when you open a Question Page in an incognito window. And I don't want to add all slugs to the default. Marked as resolved for now.

apps/client/src/pages/LocationPage.js
robinpiets: Good point. I was thinking about this as well. But the PR is already going out of control. I will mark this as a separate ticket.

apps/client/src/pages/LocationPage.js
robinpiets: Solved

apps/client/src/pages/QuestionsPage.js
robinpiets: Code is removed

apps/client/src/pages/QuestionsPage.js
robinpiets: I have no idea, I did not touch this. I will put it back for now. 

@svenjens do you know?

robinpiets: Fixed: put it back

apps/client/src/pages/QuestionsPage.js
robinpiets: `if (question.options && value !== undefined) {` ?
Or shall we move this to a different ticket?

robinpiets: Replacing this with value !== undefined is working. Marking as resolved

apps/client/src/pages/QuestionsPage.js
robinpiets: On a 'No' answer the value in console.log is `no` and not `false`

apps/client/src/pages/QuestionsPage.js
robinpiets: Great!

apps/client/src/pages/QuestionsPage.js
robinpiets: solved

apps/client/src/pages/ResultsPage.js
robinpiets: Nice catch

apps/client/src/pages/ResultsPage.js
robinpiets: Nice catch, linter did not fix this

apps/client/src/sttr_client/models/checker.js
robinpiets: I had problems with infinite loops. After I added `|| prevId === this._last.id` it seems to have been solved. But I still wanted to keep this as a backup because while (done === false) { is very dangerous.

Remove this fallback?

apps/client/src/hoc/withAddress.js
robinpiets: Fixed

apps/client/src/sttr_client/models/checker.js
afjlambert: Yes, let's remove it.

afjlambert (approved):

apps/client/src/sttr_client/models/checker.js
robinpiets: Removed (-:

PR #107, opened 2020-06-03
Demo saveCheckerResult function

robinpiets:
### Pseudo code alert!

Draft of a `saveCheckerResult` function

this is the temporary result:

```
{
  "environmentUrl": "http://localhost:3000/dakraam-plaatsen/conclusie",
  "appVersion": "1.3.0",
  "topic": {
    "slug": "dakraam-plaatsen",
    "sttrFile": "dakraam.json",
    "text": {
      "heading": "Vergunningcheck dakraam plaatsen",
      "locationIntro": "Voer het adres in waar u het dakraam wilt gaan plaatsen",
      "addressPage": "Gaat u meer dan 1 dakraam plaatsen? Doe dan per dakraam de vergunningcheck."
    },
    "intro": "DakraamIntro"
  },
  "activities": [{
    "id": "monument activity",
    "version": "1.2.4"
  }, {
    "id": "building activity",
    "version": "1.3.2"
  }],
  "questionAnswers": {
    "uitv__a165828d-8708-4db6-92c4-9c767ad7cc72": false,
    "uitv__db3e3c46-1efc-404a-8520-94974c5a1061": true,
    "uitv__ee8c045e-b188-4285-8a60-f6b115f692e1": true,
    "uitv__13eea7b4-8965-4a08-9e58-53b7ce506682": false,
    "uitv__e27d16fb-f4dd-496a-9212-8550153875e4": false,
    "uitv__461052be-8046-4d9e-9c0d-c4989650e765": true
  },
  "permitConclusions": [{
    "outcome": "\"Toestemmingsvrij\"",
    "title": " Het wijzigen van een monument bij een dakraam: Toestemmingsvrij",
    "description": "U woont niet in een monument en hebt geen vergunning nodig voor het onderdeel monument."
  }, {
    "outcome": "\"Vergunningplicht\"",
    "title": " Dakraam bouwen: Vergunningplicht",
    "description": "U hebt een vergunning nodig voor het onderdeel bouwen. U vraagt deze vergunning aan bij het landelijke omgevingsloket. Klik hiervoor op onderstaande knop."
  }],
  "registerLookups": {
    "bag": {
      "__typename": "Address",
      "id": "MDM2MzAxMDAxMjA2MjA2NA==",
      "streetName": "Louise de Colignystraat",
      "postalCode": "1055XD",
      "houseNumber": 19,
      "houseNumberFull": "19 C",
      "residence": "Amsterdam",
      "restrictions": [{
        "__typename": "Monument",
        "name": "Gemeentelijk monument"
      }],
      "zoningPlans": [{
        "name": "Paraplubestemmingsplan Stadsdeel West",
        "__typename": "zoningPlan"
      }, {
        "name": "Landlust en Gibraltarbuurt",
        "__typename": "zoningPlan"
      }, {
        "name": "Aanpassen geluidzone Westpoort en Hoogtij",
        "__typename": "zoningPlan"
      }, {
        "name": "Drijvende bouwwerken",
        "__typename": "zoningPlan"
      }]
    }
  }
}
```

### DIFF
--- a/apps/client/src/pages/ConclusionPage.js
+++ b/apps/client/src/pages/ConclusionPage.js
@@ -99,6 +99,104 @@ const ConclusionPage = ({ topic, checker }) => {
   };
   const currentDateTime = date.toLocaleDateString("nl-NL", options);
 
+  /**
+   * Demo saveCheckerResult function
+   *
+   * This function is called:
+   * - every time the conclusion page is loaded (without user data)
+   * - every time the user wants to save the progress (including user data)
+   * - every time the user wants to save the conclusion (including user data)
+   *
+   * This function is called from another file. This is just a sketch.
+   */
+  const saveCheckerResult = () => {
+    const mutationData = {
+      environmentUrl: window.location.href,
+      appVersion: process.env.REACT_APP_VERSION,
+      topic,
+
+      // How in the world are we going to get STTR version(s)?
+      // In topic.sttrFile we know the JSON file, but that's it
+      // If we really want the STTR version, we can do that 2 ways:
+      // 1) We go into topics.source.json and go get all activity id's and with those go into the xml files
+      // 2) (Re)decide to add them into the [topic].json file
+      activities: [
+        {
+          id: "monument activity",
+          version: "1.2.4",
+        },
+        {
+          id: "building activity",
+          version: "1.3.2",
+        },
+      ],
+
+      // Get `questionAnswers()` from checker.js in branch feature/autofill PR#127
+      // We need our own unique question id's (questionHash?) and not the changing id's from FL
+      // `getQuestionAnswers()` example array:
+      // [
+      //   'chappie-unique-hash-839472340': "Yes, because",
+      //   'chappie-unique-hash-839472341': false,
+      //   'chappie-unique-hash-839472342': "No, because",
+      //   'chappie-unique-hash-839472343': true,
+      // ]
+      // questionAnswers: getQuestionAnswers(),
+
+      // The current `displayConclusions()` needs a rename and a small refactor
+      // Example array:
+      // [
+      //   {
+      //     outcome: "Toestemmingsvrij",
+      //     title: "Het wijzigen van een monument bij een dakraam",
+      //     description: "U woont...",
+      //   },
+      //   {
+      //     outcome: "Vergunningplicht",
+      //     title: "Dakraam bouwen",
+      //     description: "U hebt...",
+      //   },
+      // ]
+      permitConclusions: displayConclusions, // needs to be refactored to permitConclusions
+
+      // See Trello card "Refactor Context"
+      registerLookups: {
+        monument: {},
+        cityScape: {},
+        bag: {
+          // We can only store information which it's imposible to relate to an individual
+          // `postalCode` and `streetname` are general information which we can store
+          postalCode: "",
+          streetname: "",
+          // All other information can only be stored when user agrees to save personal data
+          houseNumber: "",
+          houseNumberFull: "",
+          id: "",
+          // ...
+        },
+      },
+
+      userInput: {
+        address: {
+          // We can only store information which it's imposible to relate to an individual
+          // `postalCode` is general information which we can store
+          postalCode,
+          // All other information can only be stored when user agrees to save personal data
+          houseNumberFull,
+
+          // suffix, // We need to store this
+        },
+
+        // Store `geo`, because it's also free of personal information
+        // geo: {
+        // ...
+        // },
+      },
+    };
+
+    console.log(mutationData);
+  };
+  saveCheckerResult();
+
   return (
     <Layout>
       <Helmet>
--- a/apps/client/src/pages/ConclusionPage.js
+++ b/apps/client/src/pages/ConclusionPage.js
@@ -140,7 +140,7 @@ const ConclusionPage = ({ topic, checker }) => {
       //   'chappie-unique-hash-839472342': "No, because",
       //   'chappie-unique-hash-839472343': true,
       // ]
-      // questionAnswers: getQuestionAnswers(),
+      questionAnswers: checker.getQuestionAnswers(),
 
       // The current `displayConclusions()` needs a rename and a small refactor
       // Example array:
@@ -156,41 +156,45 @@ const ConclusionPage = ({ topic, checker }) => {
       //     description: "U hebt...",
       //   },
       // ]
-      permitConclusions: displayConclusions, // needs to be refactored to permitConclusions
+      permitConclusions: displayConclusions, // needs to be refactored and renamed to permitConclusions
 
-      // See Trello card "Refactor Context"
+      // Temp fix until autofill is merged with develop
       registerLookups: {
-        monument: {},
-        cityScape: {},
-        bag: {
-          // We can only store information which it's imposible to relate to an individual
-          // `postalCode` and `streetname` are general information which we can store
-          postalCode: "",
-          streetname: "",
-          // All other information can only be stored when user agrees to save personal data
-          houseNumber: "",
-          houseNumberFull: "",
-          id: "",
-          // ...
-        },
+        bag: sessionContext.address?.[slug],
       },
+      // But. We need to refactor this.
+      // See Trello card "Refactor Context"
+      // registerLookups: {
+      //     monument: {},
+      //     cityScape: {},
+      //     bag: {
+      //       // We can only store information which it's imposible to relate to an individual
+      //       // `postalCode` and `streetname` are general information which we can store
+      //       postalCode: "",
+      //       streetname: "",
+      //       // All other information can only be stored when user agrees to save personal data
+      //       houseNumber: "",
+      //       houseNumberFull: "",
+      //       id: "",
+      //       // ...
+      //     },
+      // },
+      // userInput: {
+      //   address: {
+      //     // We can only store information which it's imposible to relate to an individual
+      //     // `postalCode` is general information which we can store
+      //     postalCode,
+      //     // All other information can only be stored when user agrees to save personal data
+      //     houseNumberFull,
 
-      userInput: {
-        address: {
-          // We can only store information which it's imposible to relate to an individual
-          // `postalCode` is general information which we can store
-          postalCode,
-          // All other information can only be stored when user agrees to save personal data
-          houseNumberFull,
+      //     // suffix, // We need to store this
+      //   },
 
-          // suffix, // We need to store this
-        },
-
-        // Store `geo`, because it's also free of personal information
-        // geo: {
-        // ...
-        // },
-      },
+      //   // Store `geo`, because it's also free of personal information
+      //   // geo: {
+      //   // ...
+      //   // },
+      // },
     };
 
     console.log(mutationData);


apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          // questionHash: why do we need `questionHash` again?
+          registryAnswer: getRegistryAnswer(question.id), // key: `registryAnswer` or `registerAnswer`
+        });
+        return acc;
+      }, []);
+    };
+
+    const mutationData = {
+      environment: "test/acc/prod", // Send from client?
+      date: new Date(), // Send from client?
```
afjlambert: Date should not be provided by the client. The GraphQL-server should add that before storing it.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
   };
   const currentDateTime = date.toLocaleDateString("nl-NL", options);
 
+  // Demo saveConclusion function
+  const saveConclusion = () => {
+    // Demo getRegistryAnswer
+    const getRegistryAnswer = (id) => {
+      // How do we get the registry answer?
+      return "true";
+    };
```
afjlambert: What we'll have with autofill is a `autofill` prop on the question with value `"monument"` for example. It also has a `answer` of course. There is no way (yet) to see if the user has overridden the register value. It also depends if we want the user to allow for that.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
   };
   const currentDateTime = date.toLocaleDateString("nl-NL", options);
 
+  // Demo saveConclusion function
```
afjlambert: Overall, this looks great!

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          // id: does `question` contain `id`? (Yes)
+          // text: does `question` contain `text`?
+          // answer: does `question` contain `answer`? // key: `answer` or `userAnswer`? (Yes)
+          // message: why do we need `message` again?
+          // questionHash: why do we need `questionHash` again?
+          registryAnswer: getRegistryAnswer(question.id), // key: `registryAnswer` or `registerAnswer`
+        });
+        return acc;
+      }, []);
+    };
```
afjlambert: I would write it like this, I think.

```
const getQuestionAndAnswers = () => this.stack.map((question) => ({
  // ...
}));
```

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+
+    const mutationData = {
+      environment: "test/acc/prod", // Send from client?
```
afjlambert: I think we don't need this. We have the environmentUrl which contains the domain (env) and the topic-slug.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      date: new Date(), // Send from client?
+
+      environmentUrl: window.location.href,
+      topic: topic,
```
afjlambert: smart to put the whole topic in here 👍

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      checkVersion: "1.0", // getCheckVerion()
```
afjlambert: Is this the version of our application or the version of the sttr-file? I think we need both.
Application version (from package.json) is available in process.env.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      // Key: `checkVersion` or `checkerVersion` or just `version`
```
afjlambert: Yeah, so applicationVersion and sttrVersion maybe?

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+    };
+
+    // Demo getQuestionAndAnswers
+    const getQuestionAndAnswers = () => {
+      return this.stack.reduce((acc, question) => {
+        acc.push({
+          ...question,
+          // id: does `question` contain `id`? (Yes)
+          // text: does `question` contain `text`?
+          // answer: does `question` contain `answer`? // key: `answer` or `userAnswer`? (Yes)
```
afjlambert: Yes, we have `answer`. I think that's a good name for it too.
Please note the value can be a list of answers (`string[]`).

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      return "true";
+    };
+
+    // Demo getQuestionAndAnswers
+    const getQuestionAndAnswers = () => {
+      return this.stack.reduce((acc, question) => {
+        acc.push({
+          ...question,
+          // id: does `question` contain `id`? (Yes)
+          // text: does `question` contain `text`?
```
afjlambert: We have the `text`, `description` and `longDescription`. I think `text` is enough.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      // How do we get the registry answer?
+      return "true";
+    };
+
+    // Demo getQuestionAndAnswers
+    const getQuestionAndAnswers = () => {
+      return this.stack.reduce((acc, question) => {
+        acc.push({
+          ...question,
+          // id: does `question` contain `id`? (Yes)
```
afjlambert: Yeah, for now this is the (frequently changing) id from FL. In the future we'll use our own id's.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+    const getRegistryAnswer = (id) => {
+      // How do we get the registry answer?
+      return "true";
+    };
+
+    // Demo getQuestionAndAnswers
+    const getQuestionAndAnswers = () => {
+      return this.stack.reduce((acc, question) => {
+        acc.push({
+          ...question,
```
afjlambert: Are you planning to remove this?

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          // id: does `question` contain `id`? (Yes)
+          // text: does `question` contain `text`?
+          // answer: does `question` contain `answer`? // key: `answer` or `userAnswer`? (Yes)
+          // message: why do we need `message` again?
```
afjlambert: No I don't think so.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          // questionHash: why do we need `questionHash` again?
```
afjlambert: Yeah, let's include it.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          registryAnswer: getRegistryAnswer(question.id), // key: `registryAnswer` or `registerAnswer`
```
afjlambert: I would go for "registerValue".

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
   };
   const currentDateTime = date.toLocaleDateString("nl-NL", options);
 
+  // Demo saveConclusion function
+  const saveConclusion = () => {
```
afjlambert: I think we are saving more then conclusions, so maybe better to name it something like `savePermitResults`, `saveCheckerResult` ?

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      // questionsAndAnswers: [{
+      //   id: '2637r87fuw-example-hash',
+      //   text: 'Question 1 title',
+      //   answer: 'false',
+      //   registryAnswer: 'true',
+      //   questionHash: '6273q87yd3ije-example-hash',
+      // }]
+
+      conclusions: conclusions,
+      // Key name `conclusions` or `permits`?
```
afjlambert: I would go for `permit` or `permitConclusions`.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      // Do we need more / specific data?
+      // Example array:
+      // permits: [{
```
afjlambert: looks good.

apps/client/src/pages/ConclusionPage.js
afjlambert: When do we call this function? Every time we show the conclusion page?

afjlambert: From where would we call this function? Did you test with mutations?

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      //   title: "Conclusion title 1",
+      //   description: "Conclusion description 1",
+      // }]
+
+      // Only store `address` when user agrees to save personal data and wants to share the conclusion
```
afjlambert: 👍

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      address: {
```
afjlambert: Maybe we should split this into user provided data and resolved data from API. That way we can check later if the address was actually correctly resolved by GraphQL.

Maybe something like this:

```
address: {
  input: {
    postalCode,
    houseNumber,
    houseNumberAddition,
  },
  resolved: {
    streetName,
    houseNumberFull,
  }
}
```
Could also be 2 separate fields, `addressInput`, `addressLookup`.

So yeah, I think it's good to make a separation between the 2.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+        streetName: streetName,
+        houseNumberFull: houseNumberFull,
+        postalCode: postalCode,
+        residence: residence, // This should always be "Amsterdam", why store this?
+      },
+    };
+  };
+
```
afjlambert, line 203: nice work!

afjlambert (changes requested):

apps/client/src/pages/ConclusionPage.js
svenjens: How do we handle skipped questions? When we save the session i think we have to save them too

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,68 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      //   title: "Conclusion title 1",
+      //   description: "Conclusion description 1",
+      // }]
+
+      // Only store `address` when user agrees to save personal data and wants to share the conclusion
+      address: {
+        streetName: streetName,
+        houseNumberFull: houseNumberFull,
+        postalCode: postalCode,
+        residence: residence, // This should always be "Amsterdam", why store this?
```
svenjens: i agree

apps/client/src/pages/ConclusionPage.js
robinpiets: Thanks!

apps/client/src/pages/ConclusionPage.js
robinpiets: I think it's even better to `concat()` with `getUpcomingQuestions()` to make it more future proof, to share halfway

apps/client/src/pages/ConclusionPage.js
robinpiets: Check

apps/client/src/pages/ConclusionPage.js
robinpiets: Do we really need both? We just decided today that we only keep the activity/sttr version in the XML files. Not in the JSON. So:
```
// How in the world are we going to get STTR version(s)?
// In topic.sttrFile we know the JSON file, but that's it
// If we really want the STTR version, we can do that 2 ways:
// 1) We go into topics.source.json and go get all activity id's and with those go into the xml files
// 2) (Re)decide to add them into the [topic].json file
```

apps/client/src/pages/ConclusionPage.js
robinpiets: Fixed in another way

apps/client/src/pages/ConclusionPage.js
robinpiets: We have all questionAnswers from `getQuestionAnswers()` (what we made for feature/refresh) which is a list if question id's and its answers. I think thats good enough, and questionText will be a bit too much maybe. I don't know what happens when question id's change in JSON updates though... What do you think?

apps/client/src/pages/ConclusionPage.js
robinpiets: @afjlambert, ah okay, so if we use our own id's, that will be great. That fixes the problem in the above comment as well. We're going to need that for this API (-:

@svenjens, we don't need to store skipped questions. With the functions we made for Refresh, we can let the checker go through the checker with provided answers and make a new route

apps/client/src/pages/ConclusionPage.js
robinpiets: If we have our own id's and not the one from FL, this is solved!

apps/client/src/pages/ConclusionPage.js
robinpiets: It's already removed

apps/client/src/pages/ConclusionPage.js
robinpiets: `registerLookups`

apps/client/src/pages/ConclusionPage.js
robinpiets: Good questions.

This function is called:
- every time the conclusion page is loaded (without user data)
- every time the user wants to save the progress (including user data)
- every time the user wants to save the conclusion (including user data)

This function is called from another file. This is just a sketch.

I did not test with mutations yet

apps/client/src/pages/ConclusionPage.js
robinpiets: I'm going to use what we discussed today:

```
userInput: {
  address: {
    postalCode: '',
    houseNumberFull: '',
  },
  // geo: {
    // ...
  // },
},
registerLookups: {
  monument: { },
  cityScape: { },

  // Only store BAG if user allows
  bag: { streetname: '' },
},
```

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+   *
+   * This function is called from another file. This is just a sketch.
+   */
+  const saveCheckerResult = () => {
+
+    const mutationData = {
+
+      environmentUrl: window.location.href,
+      appVersion: process.env.REACT_APP_VERSION,
+      topic: topic,
```
afjlambert: Pls remove`: topic`

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+
+      // How in the world are we going to get STTR version(s)?
+      // In topic.sttrFile we know the JSON file, but that's it
+      // If we really want the STTR version, we can do that 2 ways:
+      // 1) We go into topics.source.json and go get all activity id's and with those go into the xml files
+      // 2) (Re)decide to add them into the [topic].json file
+      activities: [
+        {
+          id: 'monument activity',
+          version: 1.2.4,
```
afjlambert: This would give a parse error. Consider making 1.2.4 a string.
Maybe good to double-check if FL is using semantic versioning.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      // We need our own unique question id's (questionHash?) and not the changing id's from FL
+      // `getQuestionAnswers()` example array:
+      // [
+      //   'chappie-unique-hash-839472340': "Yes, because",
+      //   'chappie-unique-hash-839472341': false,
+      //   'chappie-unique-hash-839472342': "No, because",
+      //   'chappie-unique-hash-839472343': true,
+      // ]
+      questionAnswers: getQuestionAnswers(),
+      // Question: Do we also want to store question text? I think that's overload.
```
afjlambert: I agree, we have the question-text in VCS/git.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+
+
+      // `getPermitConclusions()` example array:
+      // [{
+      //   title: "Conclusion title 1",
```
afjlambert: Here it would actually say if you need a permit or not. Maybe good to note in this comment.

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+      registerLookups: {
+        monument: { },
+        cityScape: { },
+
+        bag: {
+          postalCode: '',
+          streetname: '',
+          // IMO, we should store all data, except for `houseNumberFull`
+          // Only store `houseNumberFull` when user agrees to save personal data
+          // All other data are free of personal information
```
afjlambert: good one!

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          houseNumberFull: '',
+          // ...
+        },
+      },
+
+      userInput: {
+
+        address: {
+          postalCode: postalCode,
+          streetName: streetName,
```
afjlambert: user does not input street name

apps/client/src/pages/ConclusionPage.js
```diff
@@ -99,6 +99,98 @@ const ConclusionPage = ({ topic, checker }) => {
...
+          // IMO, we should store all data, except for `houseNumberFull`
+          // Only store `houseNumberFull` when user agrees to save personal data
+          // All other data are free of personal information
+          houseNumberFull: houseNumberFull,
```
afjlambert: maybe split house number + addition because that's what the user actually inputs.

afjlambert (approved):

apps/client/src/pages/ConclusionPage.js
robinpiets: Any feedback on "How in the world are we going to get STTR version(s)?"

apps/client/src/pages/ConclusionPage.js
robinpiets: Good catch!

apps/client/src/pages/ConclusionPage.js
robinpiets: good catch

svenjens (approved):

robinpiets:
Closing this PR. Future feature. Still a ticket open in Trello.

PR #151, opened 2020-06-15
fixed the intro kozijnen image scale

svenjens:
for object fit to work, we need a width and a height. 
The max-width is needed for desktop


robinpiets:
Tested in IE11? 

image

robinpiets (approved):

PR #153, opened 2020-06-16
Improve eslint configuration, fix some warnings

afjlambert:
I found that the build server can fail when we run `react-scripts build`. This is because CRA does not allow for warnings during the build. I think this is a nice feature. This PR enables the same (kind of) behavior in de precommit hook.

Whats strange to me is why `--fix` did not fix these unused imports, but that's another question/issue.

I also fixed the warnings that were in the `apps/client`. The `apps/graphql` did not only have warnings but a lot of errors too. This should be fixed in a separate task.

A piece of code was uncommented because it's actually a good idea to implement that.

Please merge.

### DIFF
--- a/apps/client/src/sttr_client/index.test.js
+++ b/apps/client/src/sttr_client/index.test.js
@@ -1,33 +1,33 @@
-import fs from "fs";
-import path from "path";
-import { promisify } from "util";
+// import fs from "fs";
+// import path from "path";
+// import { promisify } from "util";
 import getChecker from ".";
 
-const readdir = promisify(fs.readdir);
+// const readdir = promisify(fs.readdir);
 
-const buildDir = path.join(__dirname, "../../.build");
+// const buildDir = path.join(__dirname, "../../.build");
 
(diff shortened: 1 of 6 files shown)


apps/client/src/sttr_client/index.test.js
```diff
@@ -1,33 +1,33 @@
...
+//   let question = checker.next();
+//   while (question) {
+//     if (question.type === "boolean") {
+//       question.setAnswer(randbool());
+//     } else {
+//       question.setAnswer("sure! why not.");
+//     }
+//     question = checker.next();
+//   }
+// };
```
robinpiets, line 30: WIP, temp or can this be removed?

robinpiets (approved):
Very nice!

svenjens (approved):

apps/client/src/sttr_client/index.test.js
afjlambert: Yeah, it's WIP but we'll do that later.

> A piece of code was uncommented because it's actually a good idea to implement that.

PR #155, opened 2020-06-17
stop tracking our events in Matamo

svenjens:


apps/client/src/components/Link.js
```diff
@@ -14,17 +15,19 @@ const Link = ({
...
   // The default action is clickExternalLink, it can be overwritten. The name in action is the topic name
   // There should always be a event name.
   const onClick = () => {
-    trackEvent({
-      category,
-      action: `${action} - ${topic.slug || "home"}`,
-      name: eventName,
-    });
+    if (trackEvents && isProduction) {
+      trackEvent({
```
afjlambert, line 25: I think it would be better if trackEvent handles `trackEvents && isProduction` by itself. It could be a noop in those cases. Maybe we need to wrap `useMatomo`'s `trackEvent`.

apps/client/src/components/Nav.js
```diff
@@ -30,18 +32,22 @@ const Nav = ({
     const action = formEnds
       ? getslug(nextText.toLowerCase())
       : "form-volgende-knop";
-    trackEvent({
-      category,
-      action,
-      name,
-    });
+    if (trackEvents && isProduction) {
```
afjlambert, line 35: Same here. It's easy to forget these checks.

apps/client/src/context.js
```diff
@@ -1,3 +1,8 @@
 import React from "react";
-export const defaultValues = { topic: null, address: null, checker: null };
+export const defaultValues = {
+  topic: null,
+  address: null,
+  checker: null,
+  trackEvents: true,
```
afjlambert, line 6: Why does this live on context and not in config?

apps/client/src/pages/DevHomePage.js
```diff
@@ -1,37 +1,41 @@
...
-            </tr>
-          ))}
-      </tbody>
-    </table>
-  </Layout>
-);
-
+const DevHomePage = () => {
+  const context = useContext(Context);
+  context.trackEvents = false;
```
afjlambert, line 12: I don't understand why this would be done if you get to DevHomePage. What if you don't visit this page? What if you visit this page half-way in your session?

afjlambert (changes requested):

PR #172, opened 2020-06-23
Refactored files to typescript

svenjens:
I'm not a typescript guru yet, so lets check this one very good.


robinpiets:
LGTM. @afjlambert ?

apps/client/src/config.ts
```diff
@@ -1,16 +1,30 @@
-export const isProduction =
+type Topic = {
+  slug: string;
```
afjlambert, line 2: This is great!

Question about these props though. How are they ordered?
I think it would be nice to have all non-optional (`?`) props listed at the top. What do you think? Could the ordre maybe even be enforced by eslint/prettier?

apps/client/src/config.ts
```diff
@@ -1,16 +1,30 @@
...
 
-export const OLO = {
+export const OLO: object = {
   home: oloHome,
   intro: `${oloHome}Particulier/particulier/home?init=true`,
   location: `${oloHome}Particulier/particulier/home/checken/LocatieWerkzaamheden`,
 };
 
-const topics = [
+const topics: Topic[] = [
```
afjlambert, line 27: 🙌 awesome!

afjlambert (approved):

apps/client/src/config.ts
svenjens: I ordered them by non-optional props on top,
Than optional props, and after that prop that are types.

robinpiets (approved):

PR #180, opened 2020-06-24
Feature/codecov GitHub action

afjlambert:


README.md
```diff
@@ -2,6 +2,9 @@
 
 This application allows residents of Amsterdam to easily see if they need a permit for a construction activity for their specific building. For various activities, like installing a window or solar panels on the roof or extending the house, we ask a number of questions about the location and plans. This leads to an outcome, either license-free or licensing obligation, where it is clear which questions lead to this conclusion. This tool can therefore be used to see how in a specific location you could carry out a construction activity without the requirement of a permit.
 
+[![codecov](https://codecov.io/gh/Amsterdam/vergunningcheck/branch/develop/graph/badge.svg)](https://codecov.io/gh/Amsterdam/vergunningcheck)
+![Github CI](https://github.com/Amsterdam/vergunningcheck/workflows/Github%20CI/badge.svg)
+
```
robinpiets: Can you put this in the bottom in the `## Thanks to` section? Or do they really require it to be in the top?

.github/workflows/codedov.yml
```diff
@@ -0,0 +1,44 @@
...
+# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
+
+name: Github CI
+
+on:
+  push:
+    branches:
+      - develop
+  pull_request:
+    branches: "*"
```
robinpiets: So this only works for develop branches, why not all of them?

robinpiets (changes requested):
It looks like it failed:
image

.github/workflows/codedov.yml
afjlambert: If you directly push to develop, yes.
But, we also let it run on all pull_requests.
Do you want to add more branches to the `push` section?

README.md
afjlambert: Positioning was odd, I agree.
Looking at other repo's in the community (like React and CRA) they add their badges to the h1 / title of the markdown. I moved it there.

afjlambert:
Not sure if Codecov 'comment' feature is not working or we need a 'base' version for it to work. Let's finish this first and then iterate if needed.

https://docs.codecov.io/docs/pull-request-comments#branches


Schermafbeelding 2020-06-24 om 16 31 17

.github/workflows/codedov.yml
robinpiets: No this is good!

codecov.yml
```diff
@@ -0,0 +1,12 @@
+codecov:
+  require_ci_to_pass: yes
+
+coverage:
+  precision: 2
+  round: down
+  range: "70...100"
```
robinpiets: Does this mean it requires 70 - 100% code coverage to pass? Lets discuss tomorrow?

.github/workflows/default.yml
```diff
@@ -0,0 +1,28 @@
+# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
robinpiets: I prefer each workflow a separate file. Let's discuss tomorrow.

robinpiets (changes requested):

codecov.yml
afjlambert: Yeah, I want it to fail on the first try. Later we can decrease this number and write more tests.

.github/workflows/default.yml
afjlambert: The downside of having separate workflow files is that we'll use much more build-minutes. Every workflow would have to do it's own, checkout, install, build.

robinpiets (approved):

PR #200, opened 2020-06-30
Image proxy

afjlambert:


apps/client/.env
```diff
@@ -1,4 +1,5 @@
 REACT_APP_VERSION=$npm_package_version
 REACT_APP_STTR_ENV=production
+REACT_APP_IMAGE_PROXY=/imageproxy/00000001002564440000/
```
svenjens, line 3: What doest this number mean?

apps/client/sttr_build/parser.js
```diff
@@ -16,6 +20,26 @@ function feelTypeMap(feel) {
...
+        if (child.type === "image") {
+          child.url = path.basename(url.parse(child.url).pathname);
+        }
+      });
+    }
+  );
+};
+
+const processMarkdown = (content) =>
+  remark().use(imgPlugin).processSync(content).toString();
```
svenjens, line 41: We used to trim processMarkdown, is this not needed anymore?

apps/imgproxy/server.js
```diff
@@ -0,0 +1,3 @@
+const app = require("./app.js");
+
+app.listen(process.env.PORT || 3003);
```
svenjens, line 3: if we have a img proxy server we can remove the imgproxy.js right?

PR #207, opened 2020-06-30
fix: Remove stored answers after checker reset

robinpiets:


svenjens (approved):

robinpiets:
PR build fails because develop is not merged with release yet. Cannot find command `build:size-report`

PR #213, opened 2020-07-01
Make bundle size smaller (replaced uniqBy and updated react-hook-form)

svenjens:


robinpiets:
Nice! `lodash.uniqby` is also in the `/apps/graphql` ;)

apps/client/src/utils/index.js
```diff
@@ -25,3 +25,26 @@ export const isObject = (val) => typeof val === "object" && val !== null;
 // Filters
 export const uniqueFilter = (value, index, self) =>
   self.indexOf(value) === index;
+
+export const uniqBy = (arr, predicate) => {
```
robinpiets, line 31: Can you add a comment?

svenjens:
> Nice! `lodash.uniqby` is also in the `/apps/graphql` ;)

But that isn't in the client bundle :-) So it isn't making our build bigger.

apps/client/package.json
```diff
@@ -25,14 +25,13 @@
...
     "lodash.isstring": "^4.0.1",
-    "lodash.uniqby": "^4.7.0",
     "named-urls": "^2.0.0",
     "react": "^16.13.1",
     "react-app-polyfill": "^1.0.6",
     "react-device-detect": "^1.12.1",
     "react-dom": "^16.13.1",
     "react-helmet": "^6.1.0",
-    "react-hook-form": "^5.7.2",
+    "react-hook-form": "^6.0.0",
```
robinpiets, line 34: Why is this in the uniqBy pr?

robinpiets (changes requested):

apps/client/package.json
svenjens: Because this is a pr to make the bundle sizes smaller. This makes the bundles size a lot smaller.

robinpiets: Cool!

robinpiets (approved):
Please change the comment and Squash and merge

PR #278, opened 2020-07-21
Fix back-back-forward bug, add DDTable for easier debuggin

afjlambert:


apps/client/src/sttr_client/models/checker.js
```diff
@@ -104,6 +104,42 @@ class Checker {
     return this.stack[this.stack.length - 1];
   }
 
+  /**
+   * find all permits that need contact,
+   * for evert permit see if the decisiveQuestion for contact
+   * equals the current (recently answered) question
```
robinpiets: 1. Can you update the comment?
```
 * Find all permits that have a "contact" conclusion,
 * for every permit see if the decisiveQuestion for contact
 * equals the current (recently answered) question
```

2. You are referring to `decisiveQuestion` which is a variable in our client, does that make sense?

apps/client/src/sttr_client/models/checker.js
```diff
@@ -104,6 +104,42 @@ class Checker {
...
+   *
+   * @returns {boolean} - if checker has a 'contact' outcome
+   * consumer can exit if wanted
+   */
+  needContactExit(currentQuestion) {
+    return !!this.permits.find((permit) => {
+      const conclusion = permit.getDecisionById("dummy");
+      const conclusionMatchingRules = conclusion.getMatchingRules();
+      const matchingContactRule = conclusionMatchingRules.find(
+        (rule) => rule.outputValue === '"NeemContactOpMet"'
```
robinpiets: maybe the stringish `'"NeemContactOpMet"'` should be a variable in the function making it useful for other systems as well?

robinpiets (approved):
Minor feedback

svenjens (approved):

PR #292, opened 2020-07-21
State management

svenjens:


afjlambert:
@svenjens Can we update the base for this RP (chappie2.0) with the latest changes from develop so it's easier to see what's actually changed? Thanks!

PR #293, opened 2020-07-21
Fix: using Memo to stop rerendering Header+Footer

robinpiets:
[If your function component renders the same result given the same props, you can wrap it in a call to `React.memo` for a performance boost in some cases by memoizing the result. This means that React will skip rendering the component, and reuse the last rendered result.](https://reactjs.org/docs/react-api.html#reactmemo)


svenjens (approved):

afjlambert (changes requested):
Can you add other components that can benefit from this? Like Header maybe?

afjlambert (approved):

PR #294, opened 2020-07-21
STTR versioning

afjlambert:


apps/client/src/sttr_client/models/permit.js
```diff
@@ -1,3 +1,5 @@
+import { isNumber } from "util";
```
robinpiets: Why use `isNumber` from `util` and not from `lodash.isnumber`, like you do in `models/question.js`? Is this different or maybe more lightweight?

apps/client/sttr_build/index.js
```diff
@@ -18,7 +18,7 @@ const outputDir = path.join(
 );
 const sttrApi = `https://sttr-builder${
   env === "PROD" ? "" : "-staging"
-}.eu.meteorapp.com/api`;
+}.eu.meteorapp.com/api/v2`;
```
robinpiets, line 21: v2 is already on production?

apps/client/sttr_build/index.js
```diff
@@ -77,31 +77,37 @@ function checkStatus(res) {
...
     const file = `${id}.json`;
     const data = {
       id,
-      permits: permits.map((permit) =>
-        sttrbuild(permitsXML.find((p) => p.id === permit.id).xml)
-      ),
+      permits: permits.map((permit) => {
+        const res = permitsConfig.find((p) => p.id === permit.id);
+        return {
+          version: res.json.version,
```
robinpiets: Could there be a case where `res.json` or `res.json.version` is `undefined`?

robinpiets (changes requested):

apps/client/sttr_build/index.js
robinpiets: Never mind

apps/client/src/sttr_client/models/permit.js
afjlambert: Good point! My IDE imported this function from `util` which I didn't even know existed. Anyway its been deprecated since node v4.0.0. I'll replace it with lodash's function.

apps/client/sttr_build/index.js
```diff
@@ -67,7 +67,7 @@ function checkStatus(res) {
       return json.map(({ _id }) => _id);
```
robinpiets, line 67: Is there any advantage to also write the version in the `topics.source.json` file?

apps/client/sttr_build/index.js
afjlambert: No there shouldn't be a reason for it to be undefined. If it does however the script should exit properly. Will double-check if that's the case. Thanks.

robinpiets (approved):

robinpiets:
Please Squash and Merge

svenjens (approved):

PR #317, opened 2020-07-23
Bug: show caption even though image loading fails

robinpiets:
Merging with version 2.0 since we have a code freeze 🥶

### DIFF
--- a/apps/client/src/components/Visual.js
+++ b/apps/client/src/components/Visual.js
@@ -23,7 +23,7 @@ export default ({ title, src, ...rest }) => {
         onLoad={() => setLoaded(true)}
         onError={handleError}
       />
-      {title && !errored && (
+      {title && (
         <figcaption>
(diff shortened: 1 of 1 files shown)


apps/client/src/components/Visual.js
```diff
@@ -23,7 +23,7 @@ export default ({ title, src, ...rest }) => {
         onLoad={() => setLoaded(true)}
         onError={handleError}
       />
-      {title && !errored && (
+      {title && (
```
svenjens, line 26: Maybe we could also handle this caption as alt in the same PR

svenjens (approved):

afjlambert (approved):

apps/client/src/components/Visual.js
robinpiets: Let's do this in another PR

PR #324, opened 2020-07-27
feature: make the stack optional

afjlambert:


robinpiets:
I really like the `isConclusive()` function. Can we merge only that function to chappie2.0? All the changes on the pages are not relevant anymore because they are going to be removed in chappie2.0

afjlambert:
You can grab it from this PR. Closing now.

PR #325, opened 2020-07-27
Chappie2.0 Wrapper / Location styling + OLO FLow

robinpiets:
#### How it looks now when active:
image

#### How it looks now when inactive:
image

----

_Text as discussed with design._

----

## Styling updates:
- Location Component copy
- Correct circle size and color to StepByStepItems
- Display correct sections
- Able to change question answers

---

### Things that need to be done in the refactor branch:
- Navigating from WrapperPage to Intro page: We need to give a warning or we need to store the checker data as well
- Detecting inside WrapperPage if the flow is either STTR or OLO, because some components only needs to be shown in STTR flow

### Things that need to be discussed with design:
- STTR Flow text (text we need to discuss because it's not in new design)
- Extra text about this activity (text that can be in both flows)

### Things that need to change in ASC-UI (PR's are open):
- Radio needs a white background, see https://github.com/Amsterdam/amsterdam-styled-components/pull/876
- Select with the `disabled` prop has an issue with background-color, see
 https://github.com/Amsterdam/amsterdam-styled-components/pull/875

### DIFF
--- a/apps/client/src/components/AddressLine.js
+++ b/apps/client/src/components/AddressLine.js
@@ -1,7 +1,12 @@
 import React from "react";
 
-export default ({ address: { streetName, houseNumberFull } }) => (
-  <strong>
-    {streetName} {houseNumberFull}
-  </strong>
-);
+export default ({ address: { streetName, houseNumberFull }, strong }) =>
+  strong ? (
+    <strong>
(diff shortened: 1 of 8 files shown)


afjlambert (approved):
No comments, looks good as far as I can judge.

PR #327, opened 2020-07-28
Feature/chappie2.0 refresh

svenjens:


afjlambert (approved):
Looks good as far as I can judge 👍

robinpiets (changes requested):
still wip right?

PR #330, opened 2020-07-28
Feature: Added zonwering STTR and Intro

robinpiets:
- Added IntroPage with (optimized) images
- Added STTR + config

An alternative for loading the images inside the App could be loading them directly from the Amazon bucket: `src="https://sttr-files.flolegal.app/00000001002564440000/Uitleg_Luik.png"`


robinpiets:
Auto label didn't work...

svenjens (approved):

svenjens:
> An alternative for loading the images inside the App could be loading them directly from the Amazon bucket: `src="https://sttr-files.flolegal.app/00000001002564440000/Uitleg_Luik.png"`


We don't have access to the bucket right? \

afjlambert:
Pushing the images to the repo works fine.
Did you try to optimize them? They seem quite large, pngcrush works great.
Code is 👌🏻

afjlambert (approved):
See comment about image file size.

robinpiets:
> See comment about image file size.

Good point!

image

PR #332, opened 2020-07-28
Feature: Added full-width background to Stepper

robinpiets:
Please squash and merge

---

## Result (update):
image

## In the design there is more spacing at the bottom than the top:
image

### DIFF
--- a/apps/client/src/routes.js
+++ b/apps/client/src/routes.js
@@ -35,14 +35,6 @@ export const routeConfig = [
     ),
   },
   {
-    exact: true,
-    name: "Stepper",
-    path: "/:slug/Stepper",
-    component: React.lazy(() =>
-      import(/* webpackPrefetch: true */ `./pages/StepperPage`)
-    ),
-  },
-  {
     name: "intro",
     exact: true,
     path: "/:slug",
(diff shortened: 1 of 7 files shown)


svenjens:
Is the margin correct? On the bottom seems to be much more space

svenjens (commented):
There seems to be more space at the bottom than on the top

apps/client/src/routes.js
```diff
@@ -34,6 +34,14 @@ export const routeConfig = [
...
     ),
   },
+  {
+    exact: true,
+    name: "Stepper",
+    path: "/:slug/Stepper",
+    component: React.lazy(() =>
+      import(/* webpackPrefetch: true */ `./pages/StepperPage`)
+    ),
+  },
```
afjlambert: Did you add this intentionally?

afjlambert (changes requested):

apps/client/src/routes.js
robinpiets: Nope. By accident. Removing it now

robinpiets:
> There seems to be more space at the bottom than on the top

The spacing had to do with the Nav component. You are free to place as much space as required.

This is the default:

image

svenjens (approved):

PR #336, opened 2020-07-29
Feat: Chappie2.0 answered questions flow

svenjens:
After changing your answer, we said the user will have to go through the question flow again.


afjlambert (approved):
Looks good.

PR #348, opened 2020-08-05
Chore: Refactor the checker.stack.map() function

robinpiets:
Refactored the `checker.stack.map()` function to make the `WrapperPage` (soon to be renamed) more sophisticated. 🎩

### DIFF
--- a/apps/client/src/components/Questions.js
+++ b/apps/client/src/components/Questions.js
@@ -2,14 +2,17 @@ import { Button, Paragraph } from "@datapunt/asc-ui";
 import React, { useContext, useState } from "react";
 
 import { SessionContext } from "../context";
-import { removeQuotes, uniqueFilter } from "../utils";
+import { removeQuotes } from "../utils";
 import Question, { booleanOptions } from "./Question";
 import { StepByStepItem } from "./StepByStepNavigation";
 
 const Questions = ({ checker, topic: { slug } }) => {
   const sessionContext = useContext(SessionContext);
   const [editQuestion, setEditQuestion] = useState(false);
-  const { questionIndex, finishedQuestions } = sessionContext[slug];
+  const { questionIndex } = sessionContext[slug];
+
+  // Styling to overwrite the line between the Items
+  const activeStyle = { marginTop: -1, borderColor: "white" };
 
   const onQuestionNext = (value) => {
     // @TODO: Let's refacter this function as well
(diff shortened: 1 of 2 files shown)


apps/client/src/components/Questions.js
```diff
@@ -10,6 +10,8 @@ const Questions = ({ checker, topic: { slug } }) => {
   const sessionContext = useContext(SessionContext);
   const [editQuestion, setEditQuestion] = useState(false);
   const { questionIndex, finishedQuestions } = sessionContext[slug];
+  // Styling to overwrite the line between the Items
+  const activeStyle = { marginTop: -1, borderColor: "white" };
```
svenjens, line 20: We have this already on chappie 2.0 right?

apps/client/src/components/Questions.js
```diff
@@ -114,68 +116,67 @@ const Questions = ({ checker, topic: { slug } }) => {
...
-    if (q.options) {
-      userAnswer = q.answer;
-    } else {
-      const responseObj = booleanOptions.find((o) => o.value === q.answer);
-      userAnswer = responseObj?.label;
-    }
+  return (
+    checker.stack
+      // Remove duplicate questions (do we really need this?)
+      .filter(uniqueFilter)
```
svenjens: We don't need this anymore, autofill is fixed.

apps/client/src/components/Questions.js
```diff
@@ -114,68 +116,67 @@ const Questions = ({ checker, topic: { slug } }) => {
...
-    return null;
-  });
+        // Define if question is the current one
+        const isCurrentQuestion =
+          q === checker.stack[questionIndex] &&
+          // isActive("questions"); @TODO needs to be added (in branch feature/refactor-stepper-logic)
+          !finishedQuestions; // @TODO needs to be removed (in branch feature/refactor-stepper-logic)
+
+        // Hide unanswered questions (eg: on browser refresh)
+        if (!isCurrentQuestion && !userAnswer) return null;
```
svenjens: We might not even need this, but lets keep it as fallback

apps/client/src/components/Questions.js
```diff
@@ -114,68 +116,67 @@ const Questions = ({ checker, topic: { slug } }) => {
...
+            customSize
+            heading={q.text}
+            highlightActive={isCurrentQuestion}
+            key={`question-${q.id}-${i}`}
+            style={isCurrentQuestion ? activeStyle : {}}
+          >
+            {isCurrentQuestion ? (
+              // Show the current question
+              <Question
+                checker={checker}
```
svenjens: this can be shortened

svenjens (changes requested):

apps/client/src/components/Questions.js
robinpiets: Yes. But since I've made it conditional I have moved it up. So it's not duplicated, but moved

apps/client/src/components/Questions.js
robinpiets: 👍

apps/client/src/components/Questions.js
robinpiets: I've double-checked it and we need it, because otherwise we get an error on refresh

apps/client/src/components/Questions.js
robinpiets: good point!

svenjens (approved):

svenjens (approved):

PR #356, opened 2020-08-06
Feat: readded prev button

svenjens:
Readded the previous button, so you can go back in your questions. 
Also i changed the onChange to save the user answer on change, so if you have given an answer it is saved while going back.

### DIFF
--- a/apps/client/src/components/Question.js
+++ b/apps/client/src/components/Question.js
@@ -23,10 +23,6 @@ export const booleanOptions = [
   },
 ];
 
-const hasKeys = (obj) =>
-  // convert to array, map, and then give the length
-  Object.entries(obj).map(([key, value]) => [key, value]).length;
-
 const Question = ({
   question: {
     id: questionId,
@@ -41,7 +37,8 @@ const Question = ({
   checker,
   editQuestion,
   setEditQuestion,
-  onSubmit: onSubmitProp,
+  onGoToNext,
+  saveAnswer,
   hideNavigation,
   showNext,
   showPrev,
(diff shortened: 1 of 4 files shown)


apps/client/src/components/Question.js
```diff
@@ -93,24 +90,16 @@ const Question = ({
...
-    // Check if data has a key that matches the questionId
-    if (
-      (onSubmitProp && !hasKeys(data)) ||
-      (hasKeys(data) && data[questionId])
-    ) {
-      onSubmitProp(data[questionId]);
-    }
+    saveAnswer(e.target.value);
+
+    if (e.target.type === "radio") setValue(e.target.name, e.target.value);
```
robinpiets: Can you add a comment?

apps/client/src/components/Questions.js
```diff
@@ -86,21 +91,25 @@ const Questions = ({
         },
       ]);
     }
+    if (questionIndex === 0) {
```
robinpiets, line 94: Can this not be `else`?

apps/client/src/pages/WrapperPage.js
```diff
@@ -110,27 +110,26 @@ const WrapperPage = ({ checker, topic, resetChecker }) => {
             <StepByStepItem
               checked={isFinished("questions")}
               customSize
-              done={isFinished("locationResult")}
+              done={isFinished("locationResult") || checker.stack.length > 0}
```
robinpiets: `checker.stack.length > 0` is equal to `checker.stack.length` But why do we need this in the first place? We can only show questions when locationResult is finished right?

robinpiets (changes requested):
Some minor feedback

apps/client/src/pages/WrapperPage.js
svenjens: We show the questions once the locationResult is finished, you can go back to change your location

apps/client/src/components/Questions.js
svenjens: This can't be else because the last question on the stack isn't necessarily the question you are editing atm

robinpiets:
Please checkout the Error in Sentry:

image

robinpiets (approved):

PR #391, opened 2020-08-12
Chore: Refactor the onQuestionNext to goToQuestion()

svenjens:
- Created a separate function to on next question

### DIFF
--- a/apps/client/src/pages/CheckerPage.js
+++ b/apps/client/src/pages/CheckerPage.js
@@ -19,6 +19,7 @@ import { geturl, routes } from "../routes";
 
 const CheckerPage = ({ checker, topic, resetChecker }) => {
   const sessionContext = useContext(SessionContext);
+  const { questionIndex } = sessionContext[topic.slug];
   const { slug, sttrFile } = topic;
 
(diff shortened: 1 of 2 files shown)


apps/client/src/pages/CheckerPage.js
```diff
@@ -66,6 +67,28 @@ const CheckerPage = ({ checker, topic, resetChecker }) => {
...
+   * @param { int | ('next'|'prev') } value - This van be, 'next, prev or a int`
+   */
+  const goToQuestion = (value) => {
+    const newIndex = Number.isInteger(value)
+      ? value
+      : value === "next"
+      ? questionIndex + 1
+      : value === "prev"
+      ? questionIndex - 1
+      : console.error(`${value} is not a int, 'next' or 'prev'`);
```
robinpiets: I like it because of its simplicity. But can you add a comment `// Return the value if value isInteger()` on line 77?

Nitpicking, but can you change the error to: `goToQuestion(): ${value} is not an integer, 'next' or 'prev'`

robinpiets (approved):
Minor suggestion for commentary

PR #394, opened 2020-08-13
Feature: Show and skip already answered questions 

robinpiets:
## Feature request from UX: 
- Always show already answered questions
- "Go to next question" should skip already answered questions

### Showing and skipping answered questions works very nice, but has the following challenges:
- Now we need a solution to force the window to go to the conclusion(eg: "scroll") otherwise the user is on the top but the conclusion is in the bottom
- How does "go to previous question" work? If this is going one question up then it's inconsistent with "go to next question", because that skips (maybe the dynamic labels (Next question / Go to conclusion) will help
- We cannot always show the EditButton on answered questions, because otherwise you might skip a question in the past which you will never be able to see

### TODO:
- [x] Fix bug "edit" button for future answered questions
- [x] Fix bug ContactConclusion also shows irrelevant getUpcoming_Questions
- [x] OLO Flow error
- Refactor a lot (in another PR)
- Feedback from UX (in another PR)

### DIFF
--- a/apps/client/src/components/QuestionAnswer.tsx
+++ b/apps/client/src/components/QuestionAnswer.tsx
@@ -6,18 +6,19 @@ import { removeQuotes } from "../utils";
 import PermitAlert from "./PermitAlert";
 
 type QuestionAnswerProps = {
+  hideEditButton?: Boolean;
   questionNeedsPermit: Boolean;
   userAnswer: string;
 };
 
 const QuestionAnswer: React.FC<
   QuestionAnswerProps & React.HTMLAttributes<HTMLElement>
-> = ({ onClick, questionNeedsPermit, userAnswer }) => {
+> = ({ hideEditButton, questionNeedsPermit, onClick, userAnswer }) => {
   return (
     <>
       <Paragraph gutterBottom={0}>
         {removeQuotes(userAnswer)}
-        <EditButton {...{ onClick }} />
+        {!hideEditButton && <EditButton {...{ onClick }} />}
       </Paragraph>
       {questionNeedsPermit && <PermitAlert />}
     </>
--- a/apps/client/src/components/Questions.js
+++ b/apps/client/src/components/Questions.js
@@ -219,10 +219,14 @@ const Questions = ({
           !q.options && booleanOptions.find((o) => o.value === q.answer);
         const userAnswer = q.options ? q.answer : booleanAnswers?.label;
 
+        // Skip unanswered questions
+        if (!userAnswer) return null;
+
         // Check if currect question is causing a permit requirement
         const questionNeedsPermit = !!permitsPerQuestion[i];
 
-        if (!userAnswer) return null;
+        // Get new index
+        const index = i + checker.stack.length;
 
         return (
           <StepByStepItem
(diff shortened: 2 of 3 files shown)


apps/client/src/components/Questions.js
```diff
@@ -66,120 +153,99 @@ const Questions = ({
...
+              // Show the answer with an edit button
+              <QuestionAnswer
+                onClick={() => onGoToQuestion(i)}
+                {...{ questionNeedsPermit, userAnswer }}
+              />
+            )}
+          </StepByStepItem>
+        );
+      })}
+      {checker._getUpcomingQuestions().map((q, i) => {
```
svenjens, line 220: Checker._getUpcomingQuestions is still private

apps/client/src/components/Questions.js
```diff
@@ -66,120 +153,99 @@ const Questions = ({
...
+        // Define userAnswer
+        const booleanAnswers =
+          !q.options && booleanOptions.find((o) => o.value === q.answer);
+        const userAnswer = q.options ? q.answer : booleanAnswers?.label;
+
+        // Skip unanswered questions
+        if (!userAnswer) return null;
+
+        // Check if currect question is causing a permit requirement
+        const questionNeedsPermit = !!permitsPerQuestion[i];
```
svenjens: Why do we need `i` here, and index below and `index` by `goToQuestion. `

robinpiets: Good point! Fixed this

apps/client/src/components/Questions.js
robinpiets: What shall we do with this? @afjlambert

svenjens (approved):

apps/client/src/components/Questions.js
afjlambert: I think you can use it this way (for now). We need to refactor sttr-client anyway.

apps/client/src/components/QuestionAnswer.tsx
```diff
@@ -6,18 +6,19 @@ import { removeQuotes } from "../utils";
 import PermitAlert from "./PermitAlert";
 
 type QuestionAnswerProps = {
+  hideEditButton?: Boolean;
   questionNeedsPermit: Boolean;
   userAnswer: string;
 };
 
 const QuestionAnswer: React.FC<
   QuestionAnswerProps & React.HTMLAttributes<HTMLElement>
```
afjlambert, line 15: I think it's weird to allow all props from html-attributes if you don't pass them down to the component.
Something like `({ hideEditButton, questionNeedsPermit, userAnswer, ...htmlProps })` and then `return <Paragraph {...htmlProps}>bla</Paragraph>`. Otherwise I'd feel comfortable providing a `onMouseOver` or `styles` for example, it would be valid according to our props but it does nothing. In this case I would add `onClick` to `QuestionAnswerProps` because other html-attributes are not supported anyway.

afjlambert: I see now it's not introduced in this PR, but still. If you have time to fix it, would be nice. I'll add it to topics to discuss.

afjlambert (approved):

apps/client/src/components/QuestionAnswer.tsx
robinpiets: `onClick` is a HTMLElement prop that I didn't want to pass as a prop. But it's good to discuss

PR #395, opened 2020-08-13
Bug: fixed issue with undefined string

robinpiets:


afjlambert (changes requested):
I don't think this improves the quality of our code. This is a util function an now does more then remove quotes.
If I pass `true` or `2` or `[]` this would still cause issues. To catch those inputs as well, this function would get more and more complicated.
Also, the function now returns `false` instead of a string. This makes the return signature different which could lead to other issues.
I would prefer converting this file to Typescript and making sure if you call removeQuotes it's actually a string.
Another solution would be calling this function like `removeQuotes(someVar || '')`.

robinpiets:
Good feedback @afjlambert! What if we make it like this:
```
export const removeQuotes = (str) => typeof str === "string" ? str.replace(/['"]+/g, "") : '';
```

afjlambert (approved):

svenjens (approved):

PR #402, opened 2020-08-17
Feature: Added ConclusionAlert for Contact Conclusion

robinpiets:
Squash and merge please


apps/client/src/components/Question.js
```diff
@@ -116,7 +117,9 @@ const Question = ({
         answers={answers}
         userAnswer={userAnswer}
       />
-      {questionNeedsPermit && <PermitAlert />}
+      {questionNeedsPermit && (
```
afjlambert: Shouldn't this be `(questionNeedsPermit || questionNeedsContactExit) &&`?

apps/client/src/components/Question.js
```diff
@@ -146,6 +149,7 @@ Question.propTypes = {
   className: PropTypes.string,
   headingAs: PropTypes.string,
   hideNavigation: PropTypes.bool,
+  questionNeedsContactExit: PropTypes.bool,
```
afjlambert, line 152: Would be nice if we could convert `Question.js` to `.ts` instead of using propTypes. Not sure how much work this is.

apps/client/src/components/QuestionAnswer.tsx
```diff
@@ -3,24 +3,33 @@ import React from "react";
...
+  userAnswer,
+}) => {
   return (
     <>
       <Paragraph gutterBottom={0}>
         {removeQuotes(userAnswer)}
         {!hideEditButton && <EditButton {...{ onClick }} />}
       </Paragraph>
-      {questionNeedsPermit && <PermitAlert />}
+      {questionNeedsPermit && (
```
afjlambert: Same here.

apps/client/src/components/Questions.js
```diff
@@ -91,20 +103,29 @@ const Questions = ({
...
 
-      goToQuestion(questionId);
-    },
-    [goToQuestion, setActiveState, setFinishedState]
-  );
+          // Set Contact Conclusion
+          setContactConclusion(true);
+
+          // End loop
+          return;
```
afjlambert: Do you mean `break`? Return doesn't end the loop.

apps/client/src/components/Questions.js
```diff
@@ -91,20 +103,29 @@ const Questions = ({
...
+        }
+      });
+    }
+  }, [checker, contactConclusion, isActive, questionIndex, onGoToQuestion, onQuestionNext, setContactConclusion, skipAnsweredQuestions, sessionContext, slug]);
```
afjlambert: 😮

apps/client/src/components/Questions.js
```diff
@@ -91,20 +103,29 @@ const Questions = ({
...
-      // Go to the specific question in the stack
-      setActiveState("questions");
-      setFinishedState(["conclusion", "questions"], false);
-      setFinishedState("locationResult", true);
+    // @TODO: Refactor this code and move to checker.js
+    // Bug fix in case of refresh: hide already future answered questions (caused by setQuestionAnswers() in withChecker)
+    if (!contactConclusion) {
+      checker.stack.forEach((q, i) => {
+        if (checker.needContactExit(q)) {
+          // Set questionIndex to question that causes the needContactExit()
```
afjlambert: Can you add the reason for setting the questionIndex in this case?

apps/client/src/components/Questions.js
```diff
@@ -181,6 +203,16 @@ const Questions = ({
...
 
+        // @TODO: Refactor this code and move to checker.js
+        // Bug fix in case of refresh: hide already future answered question (caused by setQuestionAnswers() in withChecker)
+        if (
+          contactConclusion &&
+          !checker._getUpcomingQuestions().length &&
+          questionIndex < i
+        ) {
+          return null;
+        }
```
afjlambert, line 223: This was a bit hard to grasp. Can you add a comment like, 'We don't want to render future questions if the current index is the decisive answer for the Contact outcome'. If my assumption is wrong, please correct ;-)

afjlambert (changes requested):

svenjens (approved):
Nice rename too

apps/client/src/components/Question.js
robinpiets: I thought the same thing when I started building it, but it turned out that if there's a Contact Conclusion that `questionNeedsPermit` is `true`. So that's why we don't need that.

apps/client/src/components/Question.js
robinpiets: Good point, I will add this to the refactor list.

apps/client/src/components/QuestionAnswer.tsx
robinpiets: Same

apps/client/src/components/Questions.js
robinpiets: `return` indeed didn't break the loop. It looked like it when I logged the result, but that turned out to be expected behaviour. 

It turns out [There's no built-in ability to break in forEach](https://stackoverflow.com/a/2641374)

apps/client/src/components/Questions.js
robinpiets: Yep... eslint

apps/client/src/components/Questions.js
robinpiets: Thanks

apps/client/src/components/Questions.js
robinpiets: You're right. Adjusted

apps/client/src/components/Questions.js
```diff
@@ -119,13 +119,21 @@ const Questions = ({
...
+    contactConclusion,
+    isActive,
+    questionIndex,
+    onGoToQuestion,
+    onQuestionNext,
+    setContactConclusion,
+    skipAnsweredQuestions,
+    sessionContext,
+    slug,
+  ]);
```
robinpiets, line 136: I didn't change anything here 🙄

apps/client/src/components/Question.js
afjlambert: Then I think the name `questionNeedsPermit` is very misleading, can you rename it please?

apps/client/src/components/QuestionAnswer.tsx
afjlambert: See above comment 😊

apps/client/src/components/Questions.js
afjlambert: So return does nothing here? Can we get rid of it, and the comment too?

apps/client/src/components/Question.js
robinpiets: Done

PR #403, opened 2020-08-18
Feat: set conclusion at active if it needs to be active. 

svenjens:
This pr takes care of conclusion beining active if something is clicked inside the conclusion component. 

It stil feels a bit light, like something is missing. Please check this


afjlambert (approved):

robinpiets (changes requested):
This can be a lot better if you add the callback function to the StepByStepItem, to enable hover and focus states. See my PR: #404

PR #411, opened 2020-08-19
Feature: Chappie2.0 route redirects to make old URLs backwards compatible

svenjens:
Instead of creating a catch all. 
This pr takes care of handling to old routes and create a permanent redirect.


afjlambert (changes requested):
We already have the concept of redirecting.

Why did you not add the following to `export const redirectConfig` in `routes.js`? 
```
["/:slug/locatie", "/:slug"],
["/:slug/adresgegevens", "/:slug"],
["/:slug/vragen", "/:slug"],
["/:slug/uitkomsten", "/:slug"],
["/:slug/conclusie", "/:slug"],
```

I do like `{ from: "/:slug/conclusie", to: "/:slug" },` signature much better then the `['', '']` I introduced before.
What would be really great is `routes.ts` with `export const redirectConfig: { from: string, to: string } = ...`.

And simply change:

```
{redirectConfig.map(([from, to]) => (
  <Redirect key={from} {...{ from, to }} />
))}
```
to
```
{redirectConfig.map((config) => (
  <Redirect key={from} {...config} />
))}
```

svenjens:
@afjlambert This is because we need the slug. I added the redirects to the others before but then we just have a redirect to "/:slug" We need a base route for the slug to work.

apps/client/src/components/Router.js
```diff
@@ -16,9 +16,20 @@ const Router = (props) => {
...
           {routeConfig
             .filter((route) => route.component)
-            .map((route, i) => (
-              <Route key={i} {...route} />
-            ))}
+            .map((route, i) => {
+              if (!route.redirect) {
+                return <Route key={i} {...route} />;
+              } else {
+                return route.redirect.map((redirect, i) => (
```
robinpiets: Here you don't need the `i`

apps/client/src/components/Router.js
```diff
@@ -16,9 +16,20 @@ const Router = (props) => {
           ))}
           {routeConfig
             .filter((route) => route.component)
-            .map((route, i) => (
-              <Route key={i} {...route} />
-            ))}
+            .map((route, i) => {
```
robinpiets: Can be refactored to (pseudo code):
```
 .map((route, i) => route.redirect ? route.redirect.map((redirect) => (<MyRedirect key={redirect.from} {...redirect} />) : <Route key={i} {...route} />)
```

robinpiets (changes requested):

afjlambert:
Can you validate if my last commit actually works?

robinpiets (approved):
LGTM!

Edit:  conflicts 😢

PR #435, opened 2020-08-24
Feature: Disable all future questions + refactored EditButton.tsx

robinpiets:
Disable all future question if the current question is last of the stack. We need this because it is very hard to detect future open questions and this is causing bugs. We can fix this by stop using the combo of checker.stack and checker._getUpcomingQuestions(). We already have a ticket for that.

## Dear reviewer: Please test carefully with the dakraam and zonnepanelen checker with multiple addresses and make sure you try to break the checker by changing future questions


apps/client/src/components/Questions.js
```diff
@@ -280,7 +292,10 @@ const Questions = ({
             key={`question-${q.id}-${index}`}
           >
             <QuestionAnswer
-              disabled={!!checker.isConclusive()}
+              disabled={
```
svenjens: Maybe this is a bit hard to read. 
Can we make this a `const disabledFutureQuestions = !!checker.isConclusive() | |  (!checker.isConclusive() && disableFutureQuestions)`

apps/client/src/components/Questions.js
```diff
@@ -203,6 +208,13 @@ const Questions = ({
         // Hide unanswered questions (eg: on browser refresh)
         if (!isCurrentQuestion && !userAnswer) return null;
 
+        // Disable all future question if this question is last of the stack
+        // We need this because it is very hard to detect future open questions and this is causing bugs
+        // @TODO: fix this by stop using the combo of checker.stack and checker._getUpcomingQuestions()
+        if (isCurrentQuestion && checker.stack.length === i + 1) {
+          disableFutureQuestions = true;
```
svenjens, line 215: Should this be a state hook? It should rerender the component right?

apps/client/src/components/Questions.js
```diff
@@ -280,7 +292,10 @@ const Questions = ({
             key={`question-${q.id}-${index}`}
           >
             <QuestionAnswer
-              disabled={!!checker.isConclusive()}
+              disabled={
+                !!checker.isConclusive() ||
```
afjlambert: It returns a bool, no need to do double negation.

afjlambert (changes requested):

apps/client/src/components/Questions.js
robinpiets: As explained this morning. Using a state hook will make this much more complicated

apps/client/src/components/Questions.js
robinpiets: Done!

apps/client/src/components/Questions.js
robinpiets: Good point!

svenjens (approved):

afjlambert (approved):

PR #463, opened 2020-09-02
Feature: disabled Matomo and replaced all event trackers 

robinpiets:
Second reviewer, please squash and merge.

- I disabled Matomo but kept the code intact to quickly re-enable
- I fixed a bug that causes almost every button to trigger a new pageview.
- `import { useMatomo } from "@datapunt/matomo-tracker-react";` is now only called inside `withTracking` 

Note: I purposely did not change the props order to prevent merge conflicts in other PR's.

### DIFF
--- a/apps/client/src/hoc/withTracking.js
+++ b/apps/client/src/hoc/withTracking.js
@@ -13,7 +13,11 @@ const withTracking = (Component) => ({ ...props }) => {
       trackPageView({});
     }
   };
-  const matomoTrackEvent = ({ category, action, name }) => {
+  const matomoTrackEvent = ({ action, category, name }) => {
+    // Temporary disable Matomo trackevents
+    return;
+
+    // eslint-disable-next-line no-unreachable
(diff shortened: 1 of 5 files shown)


apps/client/src/hoc/withTracking.js
```diff
@@ -13,7 +13,11 @@ const withTracking = (Component) => ({ ...props }) => {
       trackPageView({});
     }
   };
-  const matomoTrackEvent = ({ category, action, name }) => {
+  const matomoTrackEvent = ({ action, category, name }) => {
+    // Temporary disable Matomo trackevents
+    return;
```
afjlambert: You could add a console.log and/or isProduction check here for easier testing and debugging.

afjlambert (approved):

apps/client/src/hoc/withTracking.js
robinpiets: smart!

svenjens (approved):

PR #475, opened 2020-09-07
Move intro pages to dynamic intro

afjlambert:
The sttr-refactor added a dynamic-intro page for non-configured topics. This intro page can be used by our configured topics as well.
The main benefits are less duplicate code, better/more testing, more similar intropages (less divergence over time).

Please make sure:
- [x] to add a label
- [ ] ~~you added the necessary documentation (either in the markdown files or inline)~~ no need because the shared component is documented and CONTRIBUTING.md is still valid.
- [ ] ~~you added the necessary automated tests.~~ The intro-components are tested thoroughly. Not sure if testing this page (basically configuration) makes sense.


apps/client/src/intros/DakraamIntro.js
```diff
@@ -1,42 +1,19 @@
...
-    <Paragraph>
-      <ContactSentence eventName={eventNames.INTRO_EXTRAORDINARY} />
-    </Paragraph>
-  </>
+  <Intro
+    usableForBullets={[
+      "het wijzigen van een monument.",
+      "het plaatsen van het dakraam zelf.",
+    ]}
+    usableForText={`Deze vergunningcheck gaat over dakramen, daklichten en lichtstraten. U
```
robinpiets, line 11: The render method for `usableForText` is not in this PR. That is already added in the `feature/sttr-refactor` branch. 

I don't understand if that's on purpose or not. Let's discuss, ok?

apps/client/src/intros/DakraamIntro.js
```diff
@@ -1,42 +1,19 @@
...
+    kunt hem gebruiken als u een nieuwe plaatst of als u een bestaande
+    vernieuwt.`}
+    exceptions={[
```
robinpiets, line 14: Same for `exceptions`

robinpiets (changes requested):

apps/client/src/intros/DakraamIntro.js
afjlambert: It's on purpose. The Intro component is introduced by sttr-refactor. Now this intro page can use that same component.

apps/client/src/intros/DakraamIntro.js
afjlambert: It's on purpose. The Intro component is introduced by sttr-refactor. Now this intro page can use that same component.

svenjens:
I think the general intropages can be deleted once we start using translation strings

robinpiets (approved):
Only lots of merge conflicts

svenjens (approved):

svenjens:
This can be merged right? @afjlambert

PR #479, opened 2020-09-08
Feat: Track individual pages instead of the container

svenjens:
Description 
We now track pageviews on individual pages instead of the defaultLayout container.
By tracking individual page we have more control over what are tracking. 

Please make sure:
- [x] to add a label


apps/client/src/components/Layouts/DefaultLayout.tsx
```diff
@@ -1,28 +1,14 @@
...
-}: DefaultLayoutProps) {
-  const { location } = useHistory();
-
-  useEffect(() => {
-    matomoPageView();
-    // @TODO: We need to fix this!
-    //eslint-disable-next-line
-  }, [location.pathname]);
-
+function DefaultLayout({ heading, children }: DefaultLayoutProps) {
```
robinpiets: If this page does nothing other than passing it's own props, then we can remove it right?

apps/client/src/pages/CheckerPage.js
```diff
@@ -70,7 +71,7 @@ const CheckerPage = ({ checker, topic, resetChecker }) => {
   };
 
   const isFinished = (component) => isActive(component, true);
-
+  matomoPageView();
```
robinpiets: This looks very weird IMO. Is this not re-rendering all the time? Should we not wrap it in a `UseEffect()`?

PR #485, opened 2020-09-10
Refactored Nav to TypeScript

svenjens:
Description 
- Refactored the nav component to typescript 
The navigation test are out of scope. 
Please squash and merge 💥

Please make sure:
- [x] to add a label


afjlambert (approved):
Looks good

apps/client/src/components/Nav.tsx
```diff
@@ -1,13 +1,23 @@
...
-const Nav = ({
+export type NavProps = {
+  formEnds: boolean;
+  nextText: string;
+  noMarginBottom: boolean;
+  onGoToNext: (event: React.FormEvent) => void;
+  onGoToPrev: (event: React.FormEvent) => void;
+  prevText: string;
+  showNext: boolean;
+  showPrev: boolean;
```
robinpiets: I think all these props are optional and now they're required. If you make them optional, you can remove the default values from `formEnds` and `noMarginBottom`.

And since there are only two default props left, you can remove the `Nav.defaultProps` and place those two in the top.

apps/client/src/components/NavStyle.ts
```diff
@@ -1,7 +1,11 @@
 import { Icon, themeColor, themeSpacing } from "@datapunt/asc-ui";
 import styled, { css } from "styled-components";
 
-export const NavStyle = styled.div`
+type NavProps = {
+  noMarginBottom?: boolean;
+};
```
robinpiets: After you made all those props optional in the previous step, you can replace this by importing NavProps from Nav (is has an export for some reason)

```
import { NavProps } from "./Nav";
```

robinpiets (changes requested):

robinpiets (approved):

PR #496, opened 2020-09-14
Fixed broken STTR files

robinpiets:
Hotfix: STTR files broke because Flo Legal updated their API without informing us 🌊


apps/client/package.json
```diff
@@ -82,7 +82,7 @@
     "fast-xml-parser": "^3.17.4",
     "import-sort-style-module": "^6.0.0",
     "jest-styled-components": "^7.0.2",
-    "lint-staged": "^10.2.9",
+    "lint-staged": "^10.3.0",
```
afjlambert: Not part of this PR.

afjlambert (changes requested):

apps/client/package.json
robinpiets: Otherwise I could not commit...

afjlambert (approved):

PR #499, opened 2020-09-14
Enable npm audit again

afjlambert:
Testing if we can use npm audit again from our internal build-server.

Please make sure:
- [x] to add a label


afjlambert:
Unfortunately this still doesn't work. I've reported this issue with internal devops. They will report back to NPM.

PR #528, opened 2020-09-22
Conclusion changes

robinpiets:
Fixed:
- Fixed bug with displaying List on Print
- Separated and styled the Disclaimer
- Fixed styles for ConclusionOutcome
- Updated text: Added period
- Updated NewCheckerModal errors
- Updated print styling
- Updated texts

Please make sure:
- [x] to add a label


apps/client/src/atoms/List.ts
```diff
@@ -6,6 +6,8 @@ type Props = {
 };
 
 export default styled(List)<Props>`
+  -webkit-print-color-adjust: exact;
```
svenjens, line 9: Do we only need this for webkit browsers ?

apps/client/src/config/index.ts
```diff
@@ -1,4 +1,4 @@
-type Topic = {
+export type Topic = {
```
svenjens: Why do we export topic here? I don't see it imported anywhere

svenjens (changes requested):

apps/client/src/components/Conclusion/NewCheckerModal.tsx
```diff
@@ -87,6 +87,7 @@ const NewCheckerModal: React.FC<{
             <Radio
               checked={saveAddress === true}
```
svenjens, line 88: I thought we said we're going to enable yes on default?

svenjens (changes requested):

apps/client/src/atoms/List.ts
robinpiets: I believe this is the only to force print the background colors )-:

apps/client/src/components/Conclusion/NewCheckerModal.tsx
robinpiets: No we leave it empty, but force the user to chose buy giving an error when leaving empty

apps/client/src/config/index.ts
robinpiets: Good point. Fixed

PR #558, opened 2020-10-01
Enabled the disabled test

svenjens:
Description 
This PR Enables the disabled tests. 

Please make sure:
- [x] to add a label

### DIFF
--- a/apps/client/src/components/RegisterLookupSummary.test.js
+++ b/apps/client/src/components/RegisterLookupSummary.test.js
@@ -1,6 +1,7 @@
 import React from "react";
 
 import addressMock from "../__mocks__/addressMock";
+import Context from "../__mocks__/context";
 import matchMedia from "../__mocks__/matchMedia";
 import { findTopicBySlug } from "../utils";
 import { EDIT_BUTTON } from "../utils/test-ids";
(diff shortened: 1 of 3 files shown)


apps/client/src/components/RegisterLookupSummary.test.js
```diff
@@ -17,20 +18,22 @@ jest.mock("react-router-dom", () => ({
...
+    return (
+      <Context addressMock={addressMock} topicMock={topic}>
+        <RegisterLookupSummary {...props} setActiveState={setActiveState} />
+      </Context>
     );
+  };
+
+  it("renders correctly in STTR Flow", () => {
+    const topicMock = "dakraam-plaatsen";
+    const topic = findTopicBySlug(topicMock);
```
robinpiets, line 35: This is defined ~twice~ thrice in the test. Would be better to place in the top. Also you can make it `const topic = findTopicBySlug("dakraam-plaatsen");` because `topicMock` is not re-used

apps/client/src/components/RegisterLookupSummary.test.js
```diff
@@ -17,20 +18,22 @@ jest.mock("react-router-dom", () => ({
...
+    const { queryByText } = render(<WrapperWithContext topic={topic} />);
```
robinpiets, line 36: Also, there's no need to pass topic here, right?

apps/client/src/components/RegisterLookupSummary.test.js
```diff
@@ -47,14 +50,35 @@ describe("RegisterLookupSummary", () => {
     });
   });
 
-  xit("renders correctly in OLO Flow", () => {
-    const setActiveState = jest.fn();
+  it("renders correctly in OLO Flow", () => {
+    const topicMock = "aanbouw-of-uitbouw-maken";
+    const topic = findTopicBySlug(topicMock);
+    const { queryByText } = render(
+      <WrapperWithContext addressFromLocation={addressMock} topic={topic} />
```
robinpiets, line 57: No need to pass `addressFromLocation={addressMock} topic={topic}` here, right?

robinpiets (approved):
Some minor details to make the file more clean. Nice that the test is working again (-:

apps/client/src/components/RegisterLookupSummary.test.js
robinpiets: Never mind. The topic should be different

apps/client/src/components/RegisterLookupSummary.test.js
robinpiets: NM

apps/client/src/components/RegisterLookupSummary.test.js
robinpiets: NM

PR #573, opened 2020-10-06
Disables the native browser auto-suggest

robinpiets:
It's very annoying that the native browser auto-suggest hovers over our new auto-suggest. This way we disable that.

Please make sure:
- [x] to add a label

### DIFF
--- a/apps/client/src/components/Location/LocationFinder.js
+++ b/apps/client/src/components/Location/LocationFinder.js
@@ -170,14 +170,13 @@ const LocationFinder = ({
 
       <ComponentWrapper>
         <LocationTextField
+          autoComplete="off" // This disables the native browser auto-suggest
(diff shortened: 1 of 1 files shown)


apps/client/src/components/Location/LocationFinder.js
```diff
@@ -170,14 +170,13 @@ const LocationFinder = ({
 
       <ComponentWrapper>
         <LocationTextField
+          autoComplete="off" // This disables the native browser auto-suggest
```
svenjens, line 173: 🙌

svenjens (approved):

apps/client/src/components/Location/LocationFinder.js
afjlambert: :'(

PR #574, opened 2020-10-06
Fixed the edit address button and refactored lots of code

robinpiets:
The Edit button for the address is now in the correct position. Also I refactored a lot of @TODO's

Please make sure:
- [x] to add a label


apps/client/src/components/RegisterLookupSummary.js
```diff
@@ -42,16 +42,18 @@ const RegisterLookupSummary = ({
     <ComponentWrapper marginBottom={hasIMTR ? "0" : null}>
       <AddressLines
         {...address}
+        editAddressRenderer={() =>
+          compact ? null : (
```
svenjens: why not do it like !compact && <ChangeAddressModal {...{ hasIMTR, setActiveState, slug }} />
imo it's easier to read. Than compact ? null :

svenjens (changes requested):

apps/client/src/components/RegisterLookupSummary.js
robinpiets: good point, fixed it

afjlambert (approved):

svenjens (approved):

PR #578, opened 2020-10-06
Fixed bugs by making app backwards compatible

robinpiets:
This PR fixes two bugs:
1. Going directly to [LOCATION_RESULT](http://localhost:3030/bouwwerk-slopen/vragen-en-conclusie) in OLO flow broke the app.
2. Refreshing the app on a finished [LOCATION_RESULT](http://localhost:3030/bouwwerk-slopen/vragen-en-conclusie) in OLO flow redirected the user back to the Location Input.
3. Enabling the `LOCATION_RESULT` as a valid active component required the app to be backwards compatible and to "remove old sections from existing local storage data"

Please make sure:
- [x] to add a label

Please squash

### DIFF
--- a/apps/client/src/pages/CheckerPage.js
+++ b/apps/client/src/pages/CheckerPage.js
@@ -24,11 +24,11 @@ import LoadingPage from "./LoadingPage";
 const CheckerPage = ({ checker, matomoTrackEvent, resetChecker, topic }) => {
   const sessionContext = useContext(SessionContext);
   const { hasIMTR, slug, text } = topic;
+
   // OLO Flow does not have questionIndex
-  const { questionIndex } = hasIMTR ? sessionContext[topic.slug] : 0;
-  const { activeComponents, answers, finishedComponents } = sessionContext[
-    slug
-  ];
+  const { questionIndex = 0 } = sessionContext[slug] || {};
+  const { activeComponents, answers, finishedComponents } =
+    sessionContext[slug] || {};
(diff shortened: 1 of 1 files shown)


apps/client/src/pages/CheckerPage.js
```diff
@@ -57,6 +58,29 @@ const CheckerPage = ({ checker, matomoTrackEvent, resetChecker, topic }) => {
...
+    // This section does not exist anymore in the IMTR flow
+    const isOldSectionActive = hasIMTR && isActive(sections.LOCATION_RESULT);
+    const newFinishedComponents = finishedComponents.filter(
+      (section) => section !== sections.LOCATION_INPUT
+    );
+
+    if (isOldSectionActive) {
+      console.warn(
+        "Resetting components, because an old section was found active"
+      );
```
afjlambert, line 72: Can we not filter out all activeComponents that no longer exists? This implementation only takes care of 1 if I'm reading this correctly.
Something in the lines of `if (!currentlyKnownComponents[activeComponent])...`.

afjlambert (approved):

apps/client/src/pages/CheckerPage.js
robinpiets: I was thinking about this, but decided to drop this since we're going to implement this in a very different way with the new state refactor next sprint.

svenjens (approved):

PR #580, opened 2020-10-06
Fixed bug in debouncing LocationNotFound by moving it to LocationFinder

robinpiets:
There was an annoying bug in console when quickly changing values in the houseNumber input. I fixed it by moving it to its parent (LocationFinder) and expanding it. This gave me more control over the debounce functionality. It's not only debouncing the `LocationNotFound` and not the `RegisterLookupSummary`, which is a good improvement.

The LocationFinder is getting massive, so we can refactor this component by separating all inputs and input handles. I'll add that improvement to the list. I'll also note to write a debounced test.

Please make sure:
- [x] to add a label

### DIFF
--- a/apps/client/src/components/Location/LocationFinder.js
+++ b/apps/client/src/components/Location/LocationFinder.js
@@ -5,6 +5,7 @@ import React, { useCallback, useEffect, useState } from "react";
 
 import { Alert, ComponentWrapper } from "../../atoms";
 import { requiredFieldText } from "../../config";
+import useDebounce from "../../hooks/useDebounce";
 import { stripString } from "../../utils";
 import { LOCATION_FOUND } from "../../utils/test-ids";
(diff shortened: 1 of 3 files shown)


apps/client/src/components/Location/LocationFinder.js
```diff
@@ -136,6 +128,15 @@ const LocationFinder = ({
   };
 
   // AutoSuggest
+  const handleAutoSuggestSelect = (option) => {
```
afjlambert, line 131: Can you deconstruct value here?

afjlambert (approved):

svenjens (approved):

apps/client/src/components/Location/LocationFinder.js
robinpiets: Done

PR #592, opened 2020-10-08
Location e2e test

afjlambert:
Added puppeteer e2e-test for new location component.
Disable `npm test` and make it localhost-only (for now) to prevent false assumption everything works, while actually testing the ux/acc env.
Added some files to make testing in puppeteer more easy.

### DIFF
--- a/apps/e2e/README.md
+++ b/apps/e2e/README.md
@@ -7,7 +7,7 @@ Maybe you also need to run `safaridriver --enable`. I'm not sure.
 
 ## Run the test-suite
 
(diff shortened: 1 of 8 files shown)


robinpiets:
Shall we change the base to `location`?

afjlambert:
Good idea @robinpiets 
I also made sure e2e is now done on localhost:3000

afjlambert:
Please squash and merge.

robinpiets (approved):

PR #602, opened 2020-10-12
Refactored some utils (added `isValidPostalcode` and tests)

robinpiets:
Refactored the utils:
- added `isValidPostalcode` to utils. 
- updated `removeQuotes` 
- removed `uniqBy` since it was unused
- added tests for `removeQuotes` and `isValidPostalcode`

Please make sure:
- [x] to add a label

### DIFF
--- a/apps/client/src/utils/index.js
+++ b/apps/client/src/utils/index.js
@@ -37,34 +37,9 @@ export const getRestrictionByTypeName = (restrictions, typeName) =>
 
 // IMTR helper
 export const removeQuotes = (str) =>
-  typeof str === "string" ? str.replace(/['"]+/g, "") : "";
+  typeof str === "string" ? str.replace(/['"]+/g, "") : str;
 
 export const addQuotes = (str) => `"${str}"`;
-
-// `uniqBy` removes duplicates from an array (of objects) and is based on lodash.uniqBy
-// see: https://stackoverflow.com/a/40808569
-export const uniqBy = (arr, predicate) => {
-  if (!Array.isArray(arr)) {
-    return [];
-  }
-
-  const cb = typeof predicate === "function" ? predicate : (o) => o[predicate];
-
-  const pickedObjects = arr
-    .filter((item) => item)
-    .reduce((map, item) => {
-      const key = cb(item);
-
-      if (!key) {
-        return map;
-      }
-
-      return map.has(key) ? map : map.set(key, item);
-    }, new Map())
-    .values();
-
-  return [...pickedObjects];
-};
 
 /**
  *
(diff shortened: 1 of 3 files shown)


svenjens (approved):

apps/client/src/utils/index.js
```diff
@@ -37,35 +37,10 @@ export const getRestrictionByTypeName = (restrictions, typeName) =>
 
 // IMTR helper
 export const removeQuotes = (str) =>
-  typeof str === "string" ? str.replace(/['"]+/g, "") : "";
+  typeof str === "string" ? str.replace(/['"]+/g, "") : str;
```
afjlambert, line 40: This does not make it easier to convert to typescript. I'm not sure why you even want to do this. It's also no a str anymore.

apps/client/src/utils/index.js
```diff
@@ -37,35 +37,10 @@ export const getRestrictionByTypeName = (restrictions, typeName) =>
...
-        return map;
-      }
-
-      return map.has(key) ? map : map.set(key, item);
-    }, new Map())
-    .values();
-
-  return [...pickedObjects];
-};
-
```
afjlambert, line 68: Why did you remove this?

apps/client/src/utils/index.test.js
```diff
@@ -85,4 +87,31 @@ describe("util", () => {
...
   });
+  test("removeQuotes", () => {
+    expect(removeQuotes('"19-c"')).toBe("19-c");
+    expect(removeQuotes('"Abc, 123. "')).toBe("Abc, 123. ");
+    expect(removeQuotes('" A-b,c.d/e{f}[g]+-h& 1 "')).toBe(
+      " A-b,c.d/e{f}[g]+-h& 1 "
+    );
+    expect(removeQuotes("text")).toBe("text");
+    expect(removeQuotes('"text')).toBe("text");
+    expect(removeQuotes(99)).toBe(99);
```
afjlambert, line 98: Hmmmm. 🤔

apps/client/src/utils/index.js
```diff
@@ -90,3 +65,16 @@ export const scrollToRef = (ref, offset = 0) =>
...
   str && str.toLowerCase().replace(/[^a-z0-9]+/gi, "");
+
+/**
+ *
+ * This function return `true` or `false` wheter or not a value is a equal to a Dutch postalcode (eg: 1055XD)
+ *
+ * @param {string} value
+ */
+export const isValidPostalcode = (value) => {
+  const postalCodeRegex = /^[1-9][0-9]{3}[\s]?[A-Za-z]{2}$/i;
```
afjlambert, line 76: I'm not really a regex guru. But the `/i` is not needed or you can replace `[A-Za-z]` with `[a-z]` I think.

afjlambert (changes requested):

apps/client/src/utils/index.js
robinpiets: `uniqBy` was unused

apps/client/src/utils/index.js
robinpiets: I agree it's not TypeScript proof, I can do that as well. But I would expect that if I run `removeQuotes(12)` that it returns `12` or `"12"`, but not `""`

afjlambert: I guess it should give a typescript error.
`12 is not a string.`

PR #605, opened 2020-10-12
Debug table git info

afjlambert:
Add git branch-name and git commit-sha to debug table. You can actually click on the link to go to Github 😊 
We need to make sure we have the data on localhost, netlify and amsterdam basis infra.
No tests implemented as this is CI-configuration and purely for dev's.

Todo:

- [ ] After merge with develop test the debug information on acceptance

It looks like this:

Schermafbeelding 2020-10-12 om 17 27 18


robinpiets:
Locally it works good. The [Netlify branch links](https://5f84743c4982040007015411--vergunningcheck.netlify.app/) seems to be going wrong. For me personally, I don't see the added value for `Commit: fd585d3`. In what case do you use that? To check the lates commit?

svenjens (approved):

afjlambert:
@robinpiets Yes, the branch 'develop' doesn't say much. The actual commit tells you which version you are looking at.
Why do you think the branch links are wrong? It links to the actual branch. In this case it's a pull request.

PR #608, opened 2020-10-13
Debug Table are TS-ed and in Accordion

robinpiets:
I've converted Debug components to TypeScript and removed them from the test coverage. Also, the `Debug Decision Table` is now in a collapsible Accordion to make the whole page more accessible.

image

Please make sure:
- [x] to add a label
- [ ] ~~you added the necessary documentation (either in the markdown files or inline)~~
- [x] you removed the components from automated tests


apps/client/src/components/HiddenDebugInfoStyles.ts
```diff
@@ -1,16 +1,14 @@
+import { themeSpacing } from "@amsterdam/asc-ui";
+/* istanbul ignore file */
 import styled from "styled-components";
 
 export const HiddenDiv = styled.div`
   display: none;
   background: #eee;
-  padding: 1em;
-
-  @media print {
```
robinpiets, line 8: It's hidden because it's always wrapped in a `HideForPrint`

svenjens (approved):

PR #610, opened 2020-10-13
Implement shared mocking package and renamed apps-folder

afjlambert:
Renamed apps-folder to 'packages' which is the lerna default.
Separated the fixture folder from graphql to a separate package and converted to typescript.
Added function to actually get an address based on it's properties, with tests.
Improved the package.json meta-data (especially the name-field).
Used the new `vergunningcheck-mocking` package from our e2e tests. Fixed a small bug because the input field name changed.


packages/e2e/package.json
```diff
@@ -26,6 +27,7 @@
     "puppeteer": "^5.3.1",
     "selenium-server": "^3.141.59",
     "ts-node": "^9.0.0",
-    "typescript": "^4.0.3"
+    "typescript": "^4.0.3",
+    "vergunningcheck-mocking": "^2.1.0"
```
afjlambert, line 31: Check this.

packages/e2e/puppeteer/tests/location-component.ts
```diff
@@ -1,10 +1,18 @@
 import { expect } from 'chai';
 import puppeteer from "puppeteer";
-import { flows, host, puppeteerOptions, address } from "../config";
+import { flows, host, puppeteerOptions } from "../config";
 import { FlowOptions } from "../types";
+import { random } from 'vergunningcheck-mocking/src/util';
+import { getFixturesByProperties, NO_MONUMENT, NO_CITY_SCAPE } from 'vergunningcheck-mocking/src/restriction';
+
+const addresses = getFixturesByProperties([NO_MONUMENT, NO_CITY_SCAPE]);
```
afjlambert, line 8: This is what it's all about.

packages/e2e/puppeteer/config.ts
```diff
@@ -10,11 +10,6 @@ export const puppeteerOptions = DEBUG ? {
   headless: false, slowMo: 150
 } : {};
 
-export const address = {
-  zipCode: '1043ap',
-  houseNumberFull: '30',
-}
-
```
afjlambert, line 17: It's gone!

packages/e2e/puppeteer/tests/location-component.ts
```diff
@@ -13,8 +21,8 @@ const testFlow = async (page: puppeteer.Page, options: FlowOptions) => {
   await page.click('[data-testid="next-button"]');
 
   await page.waitForSelector('[name="postalCode"]');
-  await page.type('[name="postalCode"]', address.zipCode);
-  await page.type('[name="houseNumber"]', address.houseNumberFull);
+  await page.type('[name="postalCode"]', address[0]);
+  await page.type('[name="houseNumberFull"]', address[1].toString());
```
afjlambert, line 25: I think we need to improve on the positional values in the future. No more `address[0]` but `address.zipCode` etc.

packages/e2e/package.json
robinpiets: This is awesome!

packages/e2e/puppeteer/tests/location-component.ts
robinpiets: Super nice!

robinpiets (approved):
Works like a charm!

packages/e2e/puppeteer/tests/location-component.ts
robinpiets: Agree

PR #636, opened 2020-10-20
Fix: not found route

svenjens:
Description 
This PR fixes a bug where it doesn't handle not founds topics correct. 

Please make sure:
- [x] to add a label

### DIFF
--- a/packages/client/src/hoc/withChecker.tsx
+++ b/packages/client/src/hoc/withChecker.tsx
@@ -8,6 +8,7 @@ import { CheckerContext, SessionContext, SessionDataType } from "../context";
 import getChecker from "../imtr_client";
 import ErrorPage from "../pages/ErrorPage";
 import LoadingPage from "../pages/LoadingPage";
+import NotFoundPage from "../pages/NotFoundPage";
 import RedirectPage from "../pages/RedirectPage";
 import topicsJson from "../topics.json";
 import { findTopicBySlug } from "../utils";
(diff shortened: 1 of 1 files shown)


packages/client/src/hoc/withChecker.tsx
```diff
@@ -117,6 +118,10 @@ export default (Component: any) => (props: Props) => {
     return resetChecker();
   }
 
+  // @TODO We should handle this in routes.
+  if (!topic) {
+    return <NotFoundPage />;
+  }
```
afjlambert, line 124: I think we don't need this logic. There is already an `if (topic) {` and in case we don't have a topic we load the page using a dynamic topic. What we need is an `error` in case the `fetch(` fails.

packages/client/src/hoc/withChecker.tsx
```diff
@@ -117,6 +118,10 @@ export default (Component: any) => (props: Props) => {
     return resetChecker();
   }
 
+  // @TODO We should handle this in routes.
```
afjlambert, line 121: I'm not sure how we should handle this in routes.

packages/client/src/hoc/withChecker.tsx
```diff
@@ -44,7 +45,7 @@ export default (Component: any) => (props: Props) => {
 
   const initChecker = async () => {
     // if the topic is not found (dynamic IMTR-checker) or the topic is found and has an imtr flow
-    if (!checker && !error && (!topic || topic.hasIMTR)) {
+    if (!checker && !error && topic && topic.hasIMTR) {
```
afjlambert, line 48: We need to also follow this path in case we don't have a topic, for dynamic imtr-checker.

afjlambert (changes requested):
Can you revert these changes and make sure we have an error in case the xhr call fails?

PR #639, opened 2020-10-21
Removed old OLO Intro files and added tests

robinpiets:
This PR removes the old OLO Intro files and adds tests for the existing Intro files. And while on it, the existing Intro files are now TS-ed

Please squash


afjlambert (approved):
🙌 

👏

PR #641, opened 2020-10-21
Upgraded components and added tests

robinpiets:
Converted some components to TS and added tests.

Score: test coverage in client goes to 56.39 (-:


packages/client/src/components/Visual.tsx
```diff
@@ -1,9 +1,16 @@
 import { captureException } from "@sentry/browser";
 import React, { useState } from "react";
 
+import { FIGCAPTION, FIGURE, IMG } from "../utils/test-ids";
 import { Caption, Figure, Img } from "./VisualStyles";
 
-export default ({ title, src, ...rest }) => {
+type VisualProps = {
+  alt?: string;
+  src?: string;
```
afjlambert: is src optional?

afjlambert (approved):
Awesome PR.

packages/client/src/components/Visual.tsx
robinpiets: Good point, updated this prop

PR #655, opened 2020-10-28
Typescript Functions refactor

robinpiets:
Refactored TypeScript. Mainly `Function` types into `() => {}`

@afjlambert, can you have a look?


svenjens (approved):

packages/client/src/components/AutoSuggestList.tsx
```diff
@@ -38,8 +43,8 @@ const StyledIcon = styled(Icon)`
 const SuggestList: React.FC<{
   activeIndex: number;
   className?: string;
-  onSelectOption: Function;
-  options: [{ id: number; value: string }];
+  onSelectOption: (arg0: Option) => void;
```
afjlambert: Why did you name this `arg0`? Why not `(option: Option)` eg.?

packages/client/src/components/AutoSuggestList.tsx
```diff
@@ -38,8 +43,8 @@ const StyledIcon = styled(Icon)`
...
+  options: [Option];
```
afjlambert: Bit off topic...
Not entirely sure but I thought `[Option]` is a list with 1 option. and `Option[]` is a list of Options. So can be more than one.

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -27,15 +28,16 @@ const isTrueExactMatch = (
...
-  setErrorMessage: Function;
-  setFocus: Function;
-  topic: any; //@TODO: Replace it with IMTR-Client's TopicType
-}> = ({
+const LocationFinder: React.FC<
+  {
+    focus: boolean;
+    sessionAddress: any; // @TODO replace any with address type.
+    setAddress: (arg0: boolean) => void;
+    setErrorMessage: (arg0: any) => void;
```
afjlambert: Would be very nice if we have an Error type here instead of any.

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -27,15 +28,16 @@ const isTrueExactMatch = (
...
+    setFocus: (arg0: boolean) => void;
```
afjlambert: arg0?

packages/client/src/components/Location/LocationSummary.tsx
```diff
@@ -18,9 +18,8 @@ import EditLocationModal from "./EditLocationModal";
 type LocationSummaryProps = {
   addressFromLocation?: any;
   isBelowInputFields?: boolean;
-  matomoTrackEvent?: Function;
-  resetChecker?: Function;
-  setActiveState?: Function;
+  resetChecker?: () => void;
+  setActiveState?: () => void;
```
afjlambert, line 22: It seemed weird to me a `setSomethingFunction` doesn't accept any arguments. So I checked and setActiveState is actually not used :-/
Which brings me to another question. How to handle unused props? We can, at least, place the props directly above the component.
What do you think?
And can you double check any non-used fields in this type?

packages/client/src/components/Modal.tsx
```diff
@@ -23,11 +23,11 @@ type ModalProps = {
...
   confirmText?: string;
-  handleConfirmButton?: Function;
-  handleOpenModal?: Function;
+  handleConfirmButton?: () => void;
+  handleOpenModal?: () => void;
   heading: string;
-  onClick?: Function;
-  openButtonRenderer?: Function;
+  onClick?: () => void;
+  openButtonRenderer?: (arg0: any) => void;
```
afjlambert: Shouldn't this be 
```
openButtonRenderer?: (props: {
    openModal: () => void;
  }) => void;
```
?

packages/client/src/hoc/withTracking.tsx
```diff
@@ -5,11 +5,21 @@ import { useParams } from "react-router-dom";
...
+  action: string;
+  category?: string;
+  name: string;
+};
+
+export type MatomoTrackEventProps = {
+  matomoTrackEvent: (arg0: MatomoEventProps) => void;
+};
+
+const withTracking = (Component: any) => ({ ...props }) => {
```
afjlambert, line 18: Can you replace any with React.FC?

packages/client/src/utils/index.ts
```diff
@@ -61,7 +61,7 @@ export const addQuotes = (str: string) => `"${str}"`;
  * @param {number} offset - pass an offset to reduce from the total distance
  */
 export const scrollToRef = (
-  ref: { current: { getBoundingClientRect: Function } },
+  ref: { current: { getBoundingClientRect: () => { top: number } } },
```
afjlambert, line 28: 🙌

afjlambert (changes requested):
Nice improvements @robinpiets !
Couple of small remarks / questions.

packages/client/src/components/Location/LocationFinder.tsx
robinpiets: Good thought, fixed it

packages/client/src/components/AutoSuggestList.tsx
robinpiets: I think you're correct. Made it `Option[]` instead of `[Option]`. Both seem to work though

packages/client/src/components/Location/LocationFinder.tsx
robinpiets: It seems you can make this whatever you want. `arg0` came from my linter: `Parameter has a name but no type. Did you mean 'arg0: boolean'?`
That's where it came from.
All these work:
 `setAddress: (arg0: boolean) => void;`
 `setAddress: (bla: boolean) => void;`
 `setAddress: (address: boolean) => void;`

robinpiets: Changed it to  `setAddress: (address: boolean) => void;` though

packages/client/src/components/Location/LocationSummary.tsx
robinpiets: I don't really comprehend this question. Let's discuss off-topic

packages/client/src/components/Modal.tsx
robinpiets: Yeah!

packages/client/src/hoc/withTracking.tsx
robinpiets: This fails. I struggled with this one.
```
Argument of type 'FC<LocationInputProps & MatomoTrackEventProps>' is not assignable to parameter of type 'FC<{}>'.
  Types of parameters 'props' and 'props' are incompatible.
    Type '{ children?: ReactNode; }' is not assignable to type 'PropsWithChildren<LocationInputProps & MatomoTrackEventProps>'.
      Type '{ children?: ReactNode; }' is missing the following properties from type 'LocationInputProps': handleNewAddressSubmit, topic
```

afjlambert: Something like this?

```

type ComponentProps = {
  matomoTrackEvent: (props: MatomoEventProps) => void;
  matomoPageView: () => void;
};

const withTracking = (Component: React.FC<ComponentProps>) => ({
  ...props
}) => {
```

PR #670, opened 2020-11-03
TypeScripted Questions + Question

robinpiets:
TypeScripted and tested Questions + Question

@afjlambert need your help with importing from the `mocking` package, it doesn't work

todo: the second mocking file (in client) can be removed once it is imported from mocking


packages/client/src/components/Questions.test.js
```diff
@@ -0,0 +1,156 @@
...
+    setAddress,
+    setFinishedState,
+  },
+};
+
+jest.mock("react-router-dom", () => ({
+  useParams: () => ({}),
+}));
+
+// @TODO: Import from i18n
```
afjlambert: ping @svenjens

packages/client/src/components/Questions.test.js
```diff
@@ -0,0 +1,156 @@
...
+
+  /**
+   *
+   * This test can be improved by testing:
+   * - goToConclusion()
+   * - setContactConclusion()
+   * - going in a different route, eg Q1>Q2>Q4>Conclusion>Q1>Q2>Q3>Q4>Q5>Conclusion
+   *
+   * */
+});
```
afjlambert: very nice tests

afjlambert (changes requested):
Yeah, let's discuss. In react-hooks-poc we have the useChecker hook. In questions you'd use useChecker so in the Question.test you wouldn't need a mock. The checker is loaded using the slug which should be mocked.

afjlambert:
As discussed, we'll wait for react-hooks-poc to land and then iterate on this PR.

afjlambert (approved):

PR #672, opened 2020-11-04
Fixed inconsistencies: TS issues, docs and a URL cleaner

robinpiets:
This PR fixes some TS issues, docs and a new function to `removeQueryStrings` from an URL.

### DIFF
--- a/packages/client/src/components/Layouts/DefaultLayout.test.tsx
+++ b/packages/client/src/components/Layouts/DefaultLayout.test.tsx
@@ -23,7 +23,7 @@ const Wrapper = ({
   formTitle,
   heading,
 }: {
-  children: React.ReactNode;
+  children: React.ReactChildren;
   formTitle: string;
   heading: string;
 }) => {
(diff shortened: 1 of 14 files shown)


packages/client/src/components/Layouts/DefaultLayout.test.tsx
```diff
@@ -23,7 +23,7 @@ const Wrapper = ({
   formTitle,
   heading,
 }: {
-  children: React.ReactNode;
+  children: React.ReactChildren;
```
afjlambert: Can you do?

```
type Props = {
  formTitle: string;
  heading: string;
}

const Wrapper: React.FC<Props> = ({
```

afjlambert: So without the `children`

afjlambert (changes requested):
Small tweak.

packages/client/src/components/Layouts/DefaultLayout.test.tsx
robinpiets: Fixed (-:

svenjens (approved):

PR #681, opened 2020-11-09
Improve flow-types

afjlambert:
This PR makes our Flow-types a bit more strict. This way it's not possible to combine `hasIMTR: true` with `redirectToOlo: true` for example.


robinpiets (approved):

svenjens (approved):

svenjens:
Smart!

PR #682, opened 2020-11-09
export getChecker as named function instead of default export

afjlambert:
Based on request by @robinpiets


afjlambert:
Please merge.

robinpiets (approved):

PR #694, opened 2020-11-16
Slopen intro

svenjens:
Description 

Please make sure:
- [ ] to add a label

### DIFF
--- a/packages/client/src/i18n/nl.ts
+++ b/packages/client/src/i18n/nl.ts
@@ -50,12 +50,9 @@ export default {
           "Deze vergunningcheck gaat over kozijnen, deuren, ramen en panelen. U kunt hem gebruiken als u nieuwe plaatst of als u bestaande vernieuwt.",
         "placing bullet": "het plaatsen van de kozijnen zelf.",
       },
-      slopen: {
-        "build without permit exception":
-          "Het gebouw, of het deel van het gebouw wat u gaat slopen, is zonder vergunning gebouwd....?",
-        "intro description":
-          "Deze vergunningcheck gaat over slopen van bouwwerken.",
-        "placing bullet": "het slopen van een bouwwerk zelf.",
+      demolition: {
+        "intro description": "Soms staat in het bestemmingsplan dat een vergunning nodig is. Deze vergunningcheck kijkt niet naar bestemmingsplannen. Nadat u alle vragen hebt beantwoord, leest u hoe u dat zelf kunt bekijken.",
+        "need report": "U wilt gaan slopen. U hebt misschien een omgevingsvergunning nodig. Het kan ook zijn dat u de sloop moet melden. Met deze vergunningcheck kunt u zien wat u moet doen. "
       },
       zonnepanelen: {
         "intro description":
--- a/packages/client/src/i18n/nl.ts
+++ b/packages/client/src/i18n/nl.ts
@@ -50,9 +50,11 @@ export default {
           "Deze vergunningcheck gaat over kozijnen, deuren, ramen en panelen. U kunt hem gebruiken als u nieuwe plaatst of als u bestaande vernieuwt.",
         "placing bullet": "het plaatsen van de kozijnen zelf.",
       },
-      demolition: {
-        "intro description": "Soms staat in het bestemmingsplan dat een vergunning nodig is. Deze vergunningcheck kijkt niet naar bestemmingsplannen. Nadat u alle vragen hebt beantwoord, leest u hoe u dat zelf kunt bekijken.",
-        "need report": "U wilt gaan slopen. U hebt misschien een omgevingsvergunning nodig. Het kan ook zijn dat u de sloop moet melden. Met deze vergunningcheck kunt u zien wat u moet doen. "
+      slopen: {
+        "intro description":
+          "Soms staat in het bestemmingsplan dat een vergunning nodig is. Deze vergunningcheck kijkt niet naar bestemmingsplannen. Nadat u alle vragen hebt beantwoord, leest u hoe u dat zelf kunt bekijken.",
+        "need report":
+          "U wilt gaan slopen. U hebt misschien een omgevingsvergunning nodig. Het kan ook zijn dat u de sloop moet melden. Met deze vergunningcheck kunt u zien wat u moet doen. ",
       },
       zonnepanelen: {
         "intro description":
--- a/packages/client/src/intros/shared/Intro.tsx
+++ b/packages/client/src/intros/shared/Intro.tsx
@@ -16,6 +16,7 @@ type Props = {
   dependantOnQuestions?: boolean;
   dependantOnSituation?: boolean;
   exceptions?: string[];
+  needReport?: boolean;
   showContactInformation?: boolean;
   usableForBullets?: string[];
   usableForText?: string;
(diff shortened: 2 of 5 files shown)


packages/client/src/intros/shared/Intro.tsx
```diff
@@ -43,14 +45,18 @@ export default ({
 
   return (
     <>
-      <Paragraph gutterBottom={usableForBullets.length > 0 ? 8 : undefined}>
-        {t("introPage.common.check for permit intro")}{" "}
-        {usableForBullets.length > 0 && t("introPage.common.permit for")}
+      { needReport && <Paragraph>
+        {t("introPage.demolition.need report")}{" "}
+      </Paragraph> }
```
robinpiets: My autocorrect is changing these lines. Also, the extra space is not necessary. `{" " }` Do you have "Format document on Save" enabled in your IDE?

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -43,14 +45,18 @@ export default ({
...
+
+      <Paragraph gutterBottom={usableForBullets?.length > 0 ? 8 : undefined}>
+        {!needReport && t("introPage.common.check for permit intro")}{" "}
+        {usableForBullets?.length > 0 && t("introPage.common.permit for")}
```
robinpiets: This is wrong, because now this `<Paragraph>` can be rendered with empty content. Can you not do:
`{needReport ? {t("introPage.demolition.need report")} ? t("introPage.common.check for permit intro")}}` ?
If this is not possible, then this paragraph needs conditional rendering as well

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -43,14 +45,18 @@ export default ({
 
   return (
     <>
-      <Paragraph gutterBottom={usableForBullets.length > 0 ? 8 : undefined}>
-        {t("introPage.common.check for permit intro")}{" "}
-        {usableForBullets.length > 0 && t("introPage.common.permit for")}
+      { needReport && <Paragraph>
+        {t("introPage.demolition.need report")}{" "}
```
robinpiets: I would change `needReport` to `demolitionChecker` (or something) because that's where the text is from.
`needReport` > `demolition.need report` looks off

robinpiets (changes requested):

packages/client/src/i18n/nl.ts
```diff
@@ -50,12 +50,9 @@ export default {
           "Deze vergunningcheck gaat over kozijnen, deuren, ramen en panelen. U kunt hem gebruiken als u nieuwe plaatst of als u bestaande vernieuwt.",
         "placing bullet": "het plaatsen van de kozijnen zelf.",
       },
-      slopen: {
-        "build without permit exception":
-          "Het gebouw, of het deel van het gebouw wat u gaat slopen, is zonder vergunning gebouwd....?",
-        "intro description":
-          "Deze vergunningcheck gaat over slopen van bouwwerken.",
-        "placing bullet": "het slopen van een bouwwerk zelf.",
+      demolition: {
```
afjlambert: All other checkers have a dutch name. I like having an English one here. Maybe add to todo we translate the others too?

packages/client/src/intros/shared/Intro.tsx
afjlambert: This is a generic intro component. Should be reusable and contain no logic for specific checkers. Translations could come from common if really needed. I think needReport is also not the best name. `outcomesIncludeNeedReport` would be better. But the text cannot contain demolition text. Another solution would be to check if the current checker has a key in the translation file, and if so, render the text here.

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -43,14 +45,18 @@ export default ({
...
+      </Paragraph> }
+
+      <Paragraph gutterBottom={usableForBullets?.length > 0 ? 8 : undefined}>
+        {!needReport && t("introPage.common.check for permit intro")}{" "}
+        {usableForBullets?.length > 0 && t("introPage.common.permit for")}
       </Paragraph>
 
-      {usableForBullets.length > 0 && (
+      {usableForBullets?.length > 0 && (
```
afjlambert, line 55: Why did you add a `?`?

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -43,14 +45,18 @@ export default ({
...
         <List data-testid={INTRO_USABLE_FOR_BULLETS} variant="bullet">
-          {usableForBullets.map((bulletText) => (
+          {usableForBullets?.map((bulletText) => (
```
afjlambert, line 57: Why did you add a `?`?

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -69,7 +75,7 @@ export default ({
         </Paragraph>
       )}
 
-      {exceptions.length > 0 && (
+      {exceptions?.length > 0 && (
```
afjlambert, line 76: Why did you add a `?`?

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -82,7 +88,7 @@ export default ({
             style={{ marginBottom: 12 }}
             variant="bullet"
           >
-            {exceptions.map((exception) => (
+            {exceptions?.map((exception) => (
```
afjlambert, line 89: Why did you add a `?`?

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -94,7 +100,7 @@ export default ({
         </>
       )}
 
-      {exceptions.length === 0 && showContactInformation && (
+      {exceptions?.length === 0 && showContactInformation && (
```
afjlambert, line 101: Why did you add a `?`?

packages/client/src/intros/shared/Intro.tsx
```diff
@@ -43,14 +45,18 @@ export default ({
...
     <>
-      <Paragraph gutterBottom={usableForBullets.length > 0 ? 8 : undefined}>
-        {t("introPage.common.check for permit intro")}{" "}
-        {usableForBullets.length > 0 && t("introPage.common.permit for")}
+      { needReport && <Paragraph>
+        {t("introPage.demolition.need report")}{" "}
+      </Paragraph> }
+
+      <Paragraph gutterBottom={usableForBullets?.length > 0 ? 8 : undefined}>
+        {!needReport && t("introPage.common.check for permit intro")}{" "}
```
afjlambert: Why don't we show this text in case the outcomes include needReport?
It can still be a combination of both right?

afjlambert (changes requested):

packages/client/src/intros/shared/Intro.tsx
afjlambert: Please add new props to this component to `Intro.test.tsx`

afjlambert:
I'm missing some context. Can you update the description (with todo's) for this Pull-request?
Maybe it's better to only open the PR if you have all the tests etc., and keep it in draft status until then?

robinpiets (approved):

robinpiets:
Merged because of deadline

PR #697, opened 2020-11-16
Implement question-collections (checkbox-support) in IMTR-client and frontend including single-checker multi-outcome

afjlambert:
Created tests + actual implementation for `collection=true` property on `Question`-model in `question.ts`.
It has various runtime checks also (like setting an Answer).

`collection=true` was implemented in the json files from the get-go, so no changes needed to IMTR-buildscript.

This PR includes single-checker multi-outcome support in imtr-client. Demo is made in frontend (packages/client) as well.

Permit order has been removed from checker.ts because the implementation was old. It was not really used in the FE anyway (except for debug-table maybe).

You can use this checker from the staging environment:

Schermafbeelding 2020-11-17 om 11 54 52


afjlambert:
Progress...
Schermafbeelding 2020-12-01 om 13 31 34
Schermafbeelding 2020-12-01 om 13 31 29

afjlambert:
I'll finish this in the current sprint.

packages/client/src/components/Question/Question.tsx
```diff
@@ -105,7 +106,7 @@ const Question: FunctionComponent<QuestionProps> = ({
         </ComponentWrapper>
       )}
 
-      <Answers {...{ errors, question, saveAnswer }} />
+      <Answers {...{ collection, errors, question, saveAnswer }} />
```
robinpiets: No need to pass `collection` because the entire `question` is passed.

packages/imtr-client/src/types.ts
```diff
@@ -3,7 +3,9 @@ import Question from "./models/question";
 
 export type Input = Question | Decision;
 export type ClientSimpleType = boolean | string | number;
-export type Answer = ClientSimpleType | undefined;
+export type AnswerType = Answer | undefined;
+export type Answer = ClientSimpleType | string[];
```
robinpiets, line 7: Why name this `AnswerType`? I think that's confusing, because I don't understand why this is a `type`. Why not just:
`export type Answer = ClientSimpleType | string[] | undefined;` 
or rename to something that's more understandable

packages/client/src/utils/index.ts
```diff
@@ -1,4 +1,5 @@
-import { ClientSimpleType, removeQuotes } from "@vergunningcheck/imtr-client";
+import { removeQuotes } from "@vergunningcheck/imtr-client";
+import { AnswerType } from "@vergunningcheck/imtr-client/src/types";
```
robinpiets: It's a default export so please import from the root.

packages/client/src/components/Answers.tsx
```diff
@@ -37,29 +37,56 @@ const Answers: FunctionComponent<AnswersProps> = ({
 
   return (
     <ComponentWrapper data-testid={QUESTION_ANSWERS}>
-      <RadioGroup name={id}>
-        {answers?.map((answer, index) => {
-          const { label, formValue, value } = answer;
+      {question.collection ? (
```
robinpiets: I've also made quite some changes to this file, can you please help me to avoid conflicts by changing this code a bit? Please first define these variables and render the Components conditionally based on the vars instead of your ternary:

```
const showCheckboxInput = collection;
const showRadioInput = !showCheckboxInput;
```

```
{showCheckboxInput && ....}
{showRadioInput && <RadioGroup name={id}>....</RadioGroup>}
```

My PR will also support `showTextInput` and in the future we can extend this with other inputs. I will also work on converting the checkboxes to support ASC-UI, in another PR.

packages/client/src/components/Question/QuestionAnswer.tsx
```diff
@@ -1,13 +1,14 @@
 import { Paragraph } from "@amsterdam/asc-ui";
-import { Answer, ClientOutcomes } from "@vergunningcheck/imtr-client";
+import type { AnswerType } from "@vergunningcheck/imtr-client";
+import { ClientOutcomes } from "@vergunningcheck/imtr-client";
```
robinpiets, line 3: It's a default export so please import from the root.

robinpiets (changes requested):
Nice work! Got some feedback based on my PR #795

packages/client/src/components/Question/Question.tsx
afjlambert: Check! Fixed.

packages/imtr-client/src/types.ts
afjlambert: An Answer is available if a question is actually answered. The AnswerType includes the option un-answerd.

afjlambert: It's to distinct between answered (Answer) and potentially un-answered (AnswerType) questions. Any suggestions for  better names?

packages/client/src/components/Question/QuestionAnswer.tsx
afjlambert: Not sure what you mean. It is imported from the root.
Please note line 2 is for importing `types` only. Line 3 is for importing actual variables.

robinpiets: Ah okay, sorry, I thought it was the same as the previous comment. 

You can keep it like this, in my PR I am converting this to:
```
import * as imtr from "@vergunningcheck/imtr-client";
```

afjlambert: Okay

afjlambert:
@robinpiets I removed the Draft status. It's not done yet as I'm struggling with the frontend implementation. Can't really test with multiple outcomes now. Maybe you can have a look at it?

packages/imtr-client/src/types.ts
robinpiets: `AnswerType` (or the original `Answer`) could be renamed to `AnsweredAnswer`, `UserAnswer`, `GivenAnswer`, `QuestionAnswer`?

packages/imtr-client/src/models/rule.ts
```diff
@@ -52,10 +52,16 @@ export default class Rule {
...
+  evaluateNew(values: AnswerType[]): number[] {
+    // XXX: `collectionOfSimpleTypes` doesn't work for the new `values`
+    // `values` can now be `[["Answer 1", "Answer 2"], ["Answer 3"]]` instead of ["Answer 1", "Answer 2", "Answer 3"]
+    // if (!collectionOfSimpleTypes(values)) {
+    //   throw Error(
+    //     `'values' should be an array of simple types, got ${JSON.stringify(
+    //       values
+    //     )}`
+    //   );
+    // }
```
robinpiets: @afjlambert Can you please make this work?

afjlambert:
I'll close this PR as we will not soon implement this feature. I'll keep the branch around.

PR #699, opened 2020-11-17
Implement question ordering

afjlambert:
Migrated last imtr stuff to typescript.
Add imtr-lib with shared types (json).
Added mock helpers for easily setting up fake checkers.
Added tests
All models now use 1 configuration object as constructors parameter, the props are exported.

Please make sure:
- [x] to add a label
- [ ] ~~you added the necessary automated tests~~ I am refactoring the test setup, that will be part of a new PR.


packages/imtr/src/transform.ts
```diff
@@ -113,27 +118,47 @@ export default async (argv: Props) => {
...
+               *   `result.prio = permit.questions.length * 10000 + question.prio`
+               * Another solution would be using the order of permit-id's in our config.
+               * See implementation below.
+               */
+              imtr.questions = imtr.questions.map((question: any) => {
+                return {
+                  ...question,
+                  prio: question.prio + count * 10000,
+                };
+              });
```
afjlambert: This is important.
We can use both ways.

Need tests for this part.

packages/imtr/src/transform.ts
```diff
@@ -79,7 +80,7 @@ export default async (argv: Props) => {
       const { slug, permits } = topic;
       const topicJsonContent = {
         permits: await Promise.all(
-          permits.map(async (permitId) => {
+          permits.map(async (permitId, count) => {
```
afjlambert: Get the order of the permit id from the config here.

packages/imtr/src/parser.ts
```diff
@@ -231,7 +231,10 @@ const getQuestions = (
     }
 
     result.id = rule.attributes.id; // TODO: generate our own hash for id's
-    if (rule[INTER_PRIORITY]) result.prio = rule[INTER_PRIORITY];
+    // Just pass prio from rule-definition here, we'll fix the prio when merging the permits together
+    result.prio = rule[INTER_PRIORITY];
```
afjlambert, line 235: I believe in our implementation (in FL) the prio field is guaranteed.

Maybe also add a test for this.

packages/imtr/src/config.ts
```diff
@@ -9,9 +9,10 @@ export const apis: APIConfig[] = [
       "dakraam-plaatsen": ["hMwHKR7Wz4FP8Dm4x", "dRy4PfDs7jQPc9gMG"],
       "kozijnen-plaatsen": ["Xm2WwYeGkNN9w6rgQ", "5CJrbgbWZP6uZsouY"],
       "bouwwerk-slopen": [
-        "bxBCdnrFZSbwxgmxC",
-        "2eYY7McaiRgfnERo9",
-        "D4mccynMtNdzYGhXF",
+        // demolition
+        "bxBCdnrFZSbwxgmxC", // monument
+        "D4mccynMtNdzYGhXF", // report
+        "2eYY7McaiRgfnERo9", // bsdg
```
afjlambert: Make sure we sort accordingly.

**Warning** ⚠️ 
**We should test all checkers again! To see if ordering still makes sense.**

packages/imtr-client/src/index.js
```diff
@@ -93,12 +93,18 @@ export const getChecker = (config) => {
       const previousByUUID = question.uuid
         ? acc.find((q) => q.uuid === question.uuid)
         : null;
+
       if (previousByUUID) {
+        if (question.prio < previousByUUID.prio) {
+          // If new question has lower prio (more important) to override the value in our accumulator
+          previousByUUID.prio = question.prio;
+        }
```
afjlambert, line 101: Always preserve lowest (most important) prio from duplicate questions. We want the first to make most sense.

packages/client/src/topics.json
```diff
@@ -19,8 +19,8 @@
       "path": "imtr/transformed/bouwwerk-slopen.json",
       "permits": [
         "bxBCdnrFZSbwxgmxC",
-        "2eYY7McaiRgfnERo9",
-        "D4mccynMtNdzYGhXF"
+        "D4mccynMtNdzYGhXF",
+        "2eYY7McaiRgfnERo9"
```
afjlambert: This order doesn't matter I think (it's generated anyway), but it's nice to see it here.

packages/imtr/src/transform.ts
svenjens: Is this the count of the amount of questions or the order, than me might need to change the name.

packages/imtr/src/config.ts
svenjens: bsdg is a dutch abbreviation. I think it is only clear for insiders.

packages/imtr/src/transform.ts
afjlambert: No it's the index of this permitId from the config, I'll rename to `permitIndex`.

packages/imtr/src/config.ts
afjlambert: thanks, fixed

packages/imtr/src/parser.ts
afjlambert: double-checked.

robinpiets (approved):

PR #711, opened 2020-11-23
The houseNumberFull input can now be used with spaces between characters

svenjens:
Description 
This PR fixes a bug where you couldn't input house numbers without spaces.

Please make sure:
- [x] to add a label

*please squash and merge*


robinpiets:
## These addresses need to work:
image

## So these combinations need to give results:
image
image
image

afjlambert:
Hi Sven can you document why you implemented it this way? Please also make sure this behavior keeps working in the future by adding the necessary unit-tests.

packages/client/src/components/Location/LocationFinder.test.js
```diff
@@ -118,6 +118,75 @@ describe("LocationFinder", () => {
...
+    await waitFor(() => screen.getByTestId(LOCATION_FOUND));
+
+    await act(async () => {
+      fireEvent.change(inputPostalCode, {
+        target: { value: "1027AE" },
+      });
+      fireEvent.change(inputHouseNumber, {
+        target: { value: "20 H L" },
+      });
+    });
```
robinpiets: Awesome tests dude! You got it working (-:

Maybe you can add come more, like:
- 1024BV, 546k 1
- 1024BV, 546k 1
- 1024BV, 54 6k 1 (not found?)

- 1027AE, 20hl
- 1027AE, 20HL
- 1027A, 20 h L
- 1027A, " 20 hl "

And maybe, if you want to be even more awesome, you can even add `matches` to the mocks and detect for each input if the AutoSuggest is open with:
- "546 K 1",
- "546 K 10",

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -63,9 +63,20 @@ const LocationFinder: React.FC<LocationFinderProps & MatomoTrackEventProps> = ({
   const [autoSuggestValue, setAutoSuggestValue] = useState<string>("");
   const [touched, setTouched] = useState<{ [key: string]: boolean }>({});
 
+  const splitHouseNumber = houseNumberFull
+    ?.toUpperCase()
+    ?.match(/[a-z]+|[^a-z]+/gi);
```
robinpiets: If i'm not mistaken you ~can~ could only use one `a-z` with this regex?

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -63,9 +63,20 @@ const LocationFinder: React.FC<LocationFinderProps & MatomoTrackEventProps> = ({
...
+
+  const graphQlHouseNumber = splitHouseNumber
+    ?.map((part) => `${part.trim()} `) // Split each part with a space
+    .toString() // Revert to one string
+    .replace(/,/g, "") // Remove all commas
+    .replace(/\s+/g, " ") // Remove duplicate spaces
+    .trim(); // Remove and spaces
```
robinpiets: This is still very messy. Possible cleanups:
- I see who `trim()` functions, is it not possible to use only one?
- I see two `replace()` functions and one `toString()`, can you not combine into one regex?
- I've created the util `stripString`, can you not use that on each `part`? That also lowercases and trims

robinpiets (changes requested):
Great job. Some cleanups

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -63,9 +63,18 @@ const LocationFinder: React.FC<LocationFinderProps & MatomoTrackEventProps> = ({
   const [autoSuggestValue, setAutoSuggestValue] = useState<string>("");
   const [touched, setTouched] = useState<{ [key: string]: boolean }>({});
 
+  const graphQlHouseNumber = houseNumberFull
```
robinpiets: Can you add a comment what you're doing here?

afjlambert: I agree, not totally clear, maybe extract as a util as well. Test it too.

afjlambert (changes requested):

packages/client/src/utils/index.test.tsx
```diff
@@ -61,6 +62,18 @@ describe("util", () => {
...
+    expect(transformHousenumberToValid("")).toBe("");
+    expect(transformHousenumberToValid("110")).toBe("110");
+    expect(transformHousenumberToValid("1 10")).toBe("1 10");
+    expect(transformHousenumberToValid("19-c")).toBe("19 - C");
+    expect(transformHousenumberToValid("19-c        ")).toBe("19 - C");
+    expect(transformHousenumberToValid("          19-c        ")).toBe(
+      "19 - C"
+    );
+    expect(transformHousenumberToValid("10hl")).toBe("10 H L");
+    expect(transformHousenumberToValid("10h l")).toBe("10 H L");
```
robinpiets: Can you add tests for:
- `&`
- `101&`
- `101 &`
- `101 &a`
- `101 &1`

Would like to see what happens with this.

packages/client/src/utils/index.test.tsx
```diff
@@ -61,6 +62,18 @@ describe("util", () => {
     expect(isValidPostalcode(" 1055XD ")).toBe(true);
     expect(isValidPostalcode(" 1055 XD ")).toBe(true);
   });
+  test("transformHousenumberToValid", () => {
```
robinpiets: We spell somthinghousenumber like `somethingHouseNumber`. Even though I'm wondering about the clarity of this function name. @afjlambert what do you think?

packages/client/src/utils/index.test.tsx
```diff
@@ -61,6 +62,18 @@ describe("util", () => {
...
+    expect(transformHousenumberToValid("")).toBe("");
+    expect(transformHousenumberToValid("110")).toBe("110");
+    expect(transformHousenumberToValid("1 10")).toBe("1 10");
+    expect(transformHousenumberToValid("19-c")).toBe("19 - C");
+    expect(transformHousenumberToValid("19-c        ")).toBe("19 - C");
+    expect(transformHousenumberToValid("          19-c        ")).toBe(
```
robinpiets: This one doesn't need so much spaces, 3 spaces on each side should be fine.

packages/client/src/utils/index.ts
```diff
@@ -95,6 +95,20 @@ export const isValidPostalcode = (value?: string) => {
...
+/**
+ *
+ * This function transforms some cases where a user for example makes a typo or makes a strange space combination to a valid option.<
+ *
+ * @param {string} value
+ */
+export const transformHousenumberToValid = (value: string) =>
+  !/[a-z]/i.test(value) // First we test if your string contains a letter. If not, return as is, there is no need processing it
+    ? value
+    : (value?.toUpperCase().match(/[^a-z\s]+|[a-z]/gi) || []) // tokenize the string into letter/non-letter & non-whitespace chunks
```
robinpiets: `/[^a-z\s]+|[a-z]/gi` is the same as `/[^A-Z\s]+|[A-Z]/g)` since you're already transformed it `toUpperCase()`.

robinpiets: Would be cleaner if you remove the `!` in the beginning of you ternary and swap the two return values.

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -66,7 +70,7 @@ const LocationFinder: FunctionComponent<LocationFinderProps> = ({
 
   const variables = {
     extraHouseNumberFull: "",
-    houseNumberFull: houseNumberFull,
+    houseNumberFull: transformHousenumberToValid(houseNumberFull),
```
robinpiets: When starting localhost I get an error `Type 'undefined' is not assignable to type 'string'.`  Don't you?

robinpiets (changes requested):
When starting I get a TS error. When I fixed this the Location Section works great! Some improvements though

packages/client/src/utils/index.ts
svenjens: I don't think so. 
That would turn out to this.

```
export const transformHouseNumberToValid = (value: string) =>
  /[a-z]/i.test(value) // First we test if your string contains a letter. If not, return as is, there is no need processing it
    ? (value?.toUpperCase().match(/[^a-z\s]+|[a-z]/gi) || []) // tokenize the string into letter/non-letter & non-whitespace chunks
        .join("") // Revert to one string
        .replace(/(?<=\D)|(?=\D)/g, " ") // Add spaces before/after a non-digit
        .trim(); // Trim space before and after
    : value

```

spikes/new-form-view.md
```diff
@@ -0,0 +1,60 @@
...
+Willen we dit implementeren op de bestaande Questions + CheckerPage components? Of maken we een nieuwe "view" die een optimale weergave voor collega's geeft?
+
+
+Willen we een losse view voor collega's? 
+ - moet dit een losse pagina zijn.  
+```
+Dit is een andere flow. Het is dus logisch hier een apparte pagina van te maken.  Componenten en functies kunnen we blijven gebruiken. 
+```
+ - Hoe gaan we om met wijzigen van vragen.
+```
```
robinpiets: Why put this in this PR?

packages/client/src/utils/index.ts
```diff
@@ -95,6 +95,24 @@ export const isValidPostalcode = (value?: string) => {
   return !!(value && value.toString().trim().match(postalCodeRegex));
 };
 
+/**
+ *
+ * This function transforms some cases where a user for example makes a typo or makes a strange space combination to a valid option.<
```
robinpiets: typo at the end of the sentence

robinpiets: Can you add examples at "some cases where a user for example makes a typo" and "makes a strange space combination to a valid option"

packages/client/src/components/Location/LocationFinder.tsx
```diff
@@ -66,7 +70,9 @@ const LocationFinder: FunctionComponent<LocationFinderProps> = ({
 
   const variables = {
     extraHouseNumberFull: "",
-    houseNumberFull: houseNumberFull,
+    houseNumberFull: houseNumberFull
+      ? sanitizeHouseNumberFull(houseNumberFull)
+      : "",
```
robinpiets: I would expect returning `""` at `sanitizeHouseNumberFull` when it's not a value. This would clean this up

robinpiets: In that case the function should not only expect `string` types. So can you also add a test for:
```
expect(sanitizeHouseNumberFull(undefined)).toBe("");
```

robinpiets (changes requested):
Nearly there

packages/client/src/utils/index.test.tsx
```diff
@@ -61,6 +62,22 @@ describe("util", () => {
     expect(isValidPostalcode(" 1055XD ")).toBe(true);
     expect(isValidPostalcode(" 1055 XD ")).toBe(true);
   });
+  test("transformHousenumberToValid", () => {
```
robinpiets: This test should be renamed to `sanitizeHouseNumberFull`

packages/client/src/utils/index.ts
```diff
@@ -95,6 +95,27 @@ export const isValidPostalcode = (value?: string) => {
...
 
+/**
+ *
+ * This function transforms some cases where a user for example makes a typo or makes a strange space combination to a valid option.
+ * For example:
+ *  10h l -> 10 H L
+ *  10hl -> 10 H L
+ * @param {string} value
+ */
+export const sanitizeHouseNumberFull = (value: string | undefined) => {
```
robinpiets: `value: string | undefined` can be `value?: string`

packages/client/src/utils/index.ts
```diff
@@ -95,6 +95,27 @@ export const isValidPostalcode = (value?: string) => {
...
+  const spacesBeforeAndAfterNonDigit = /(?<=\D)|(?=\D)/g;
+  const nonDigits = /[a-z]/i;
+  const tokenizeLetterAndNonLetter = /[^A-Z\s]+|[A-Z]/g;
+
+  if (!value) return "";
+  if (!nonDigits.test(value)) return value;
+  return (value?.toUpperCase().match(tokenizeLetterAndNonLetter) || []) // tokenize the string into letter/non-letter & non-whitespace chunks
```
robinpiets: The question-mark is unnecessary here, so it can be `return (value.toUpperCase().match(tokenizeLetterAndNonLetter) || [])`

robinpiets (approved):

robinpiets:
@afjlambert Can you review as well?

afjlambert (approved):

PR #714, opened 2020-11-25
ClientOutcomes is more robust because it's converted to enum

robinpiets:
This PR made `ClientOutcomes` way more robust by converting it to enum. Also, I've fixed the bug with the missing outcome text.

Please squash when approved.

Please make sure:
- [x] to add a label

### DIFF
--- a/packages/client/src/components/Questions.tsx
+++ b/packages/client/src/components/Questions.tsx
@@ -1,10 +1,10 @@
 import { setTag } from "@sentry/browser";
 import {
   Checker,
+  ClientOutcomes,
   Decision,
   Question as ImtrQuestion,
   Permit,
-  clientOutcomes,
   imtrOutcomes,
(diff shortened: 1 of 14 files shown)


svenjens (approved):

packages/client/src/components/Questions.tsx
```diff
@@ -344,7 +345,7 @@ const Questions: React.FC<
         const showQuestionAlert = !!permitsPerQuestion[i];
 
         // Define the outcome type
-        const outcomeType = permitsPerQuestion[i];
+        const outcomeType = permitsPerQuestion[i] as ClientOutcomes;
```
afjlambert: Can you use : instead of as?

afjlambert (changes requested):

afjlambert (approved):

PR #715, opened 2020-11-26
Fixed bug with multiple pageviews

robinpiets:
This PR fixes the annoying bug with multiple pageview events on reload. The problem was that the LoadingPage was also recording pageviews. This should not happen, because it's not really a 'page'. It just uses the Topic Layout to look nice.


svenjens (approved):

robinpiets:
Merging this @afjlambert so I can continue debugging the other bug. If you have feedback I will change on the other PR

PR #716, opened 2020-11-26
Fixed custom-hooks bugs and added more TypeScript

robinpiets:
This PR fixes:
- the annoying bug with irregularities in getting / setting topicData. The problem was only in `packages/client/src/SessionContext.tsx` and had to do with not updating when the slug changed.
- failed autofill when `checker` was already set
- TypeScript issues with `Address`


packages/client/src/components/Router.tsx
```diff
@@ -1,13 +1,13 @@
...
+import { Redirect, Route, Switch } from "react-router-dom";
 
 import LoadingPage from "../pages/LoadingPage";
 import { redirectConfig, routeConfig } from "../routes";
 import ScrollToTop from "./ScrollToTop";
 
 const Router = () => {
   return (
-    <BrowserRouter>
+    <>
```
afjlambert: Maybe add a comment that browser router is actually defined somewhere else. This could be confusing.

packages/client/src/hooks/useSlug.ts
```diff
@@ -5,7 +5,7 @@ import { getSlugFromPathname } from "../utils";
 export default (): string => {
   const { pathname } = useLocation();
   const slug = getSlugFromPathname(pathname);
-  if (!slug) {
+  if (!slug && pathname !== "/") {
```
afjlambert, line 8: if pathname === "/" then there is no slug. Why add this check here?

packages/client/src/index.tsx
```diff
@@ -38,19 +39,21 @@ const AppGlobalStyle = createGlobalStyle`
...
-          <GlobalStyle />
-          <AppGlobalStyle />
-          <MatomoProvider value={createInstance(matomo)}>
-            <Router />
-          </MatomoProvider>
-        </ThemeProvider>
-      </ApolloProvider>
-    </CheckerProvider>
-  </SessionProvider>,
+  <BrowserRouter>
```
afjlambert, line 42: great find!

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -14,16 +14,26 @@ import {
...
+import ErrorPage from "./ErrorPage";
 import LoadingPage from "./LoadingPage";
 
 const CheckerPage = () => {
   const { topicData, setTopicData } = useTopicData();
   const { checker } = useChecker();
   const topic = useTopic();
+  const slug = useSlug();
   const { matomoTrackEvent } = useTracking();
   const { text } = topic;
```
afjlambert: You can destructure slug and text directly from useTopic?
`const {slug, text} = useTopic();`
No need for useSlug

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -111,6 +120,10 @@ const CheckerPage = () => {
     return <LoadingPage />;
   }
 
+  if (!sessionContext.session[slug]) {
```
afjlambert, line 124: Why not test for topicData instead? Importing sessionContext is not needed then?

packages/client/src/utils/test-utils.tsx
```diff
@@ -75,17 +76,19 @@ const TestProvider: React.FC<TestProviderProps> = ({
...
-      <ApolloProvider client={getTestClient(mocks)}>
-        <ThemeProvider>
-          <MatomoProvider value={createInstance(matomo)}>
-            {children}
-          </MatomoProvider>
-        </ThemeProvider>
-      </ApolloProvider>
-    </CheckerProvider>
-  </SessionProvider>
+  <BrowserRouter>
```
afjlambert, line 79: check

afjlambert (changes requested):
Looks great, some minor points.
(edited; oh I see it's still draft)

svenjens (approved):

packages/client/src/hooks/useSlug.ts
robinpiets: The DevHomePage has no slug and is not accessible because of this `throw Error()`. Or shall I remove this check for `!slug`?

packages/client/src/pages/CheckerPage.tsx
robinpiets: Because the whole problem is that topicData can be defined and holds the data of another slug. That's why this check is needed.

robinpiets: I will add a comment

robinpiets: Or when someone manually deletes the Session Storage

packages/client/src/pages/CheckerPage.tsx
robinpiets: 1. I don't know why, but I cannot get slug from useTopic: the bug with `topicData` returns. I **need** to get slug here from `useSlug`. ❌ 
2. I can destruct `text` from `useTopic` ✅

packages/client/src/SessionContext.tsx
```diff
@@ -2,11 +2,37 @@ import { Answer } from "@vergunningcheck/imtr-client";
...
+export type Restriction = {
+  __typename?: string;
+  name?: string;
+  scope?: string;
+};
+export type ZoningPlan = {
+  __typename?: string;
+  name?: string;
+  scope?: string;
+};
```
robinpiets, line 17: How to make this prettier? @afjlambert

packages/client/src/SessionContext.tsx
```diff
@@ -80,19 +106,30 @@ export const SessionProvider: React.FC = ({ children }) => {
...
-  defaultTopicSession.type = slug;
+  const slug = useSlug();
 
   const initialState = session[slug] || defaultTopicSession;
   const [topicData, setTopicData] = useReducer(topicReducer, initialState);
 
   useEffect(() => {
-    sessionStorage.setItem("sessionData", JSON.stringify(session));
-  }, [session]);
+    // Update the `session` with default `topicData` when the `slug` changes
```
robinpiets, line 115: These useEffects solve the bug with the inconsistent `topicData`

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -158,12 +172,15 @@ const CheckerPage = () => {
...
-  const handleNewAddressSubmit = (address: any) => {
+  const handleNewAddressSubmit = (address: Address) => {
     setTopicData({
       activeComponents: [sections.QUESTIONS],
       finishedComponents: [sections.LOCATION_INPUT],
       address,
     });
+
+    // Autofill `checker` when `address` is submitted
+    checker.autofill(autofillResolvers, { address });
```
robinpiets, line 183: Since `checker` is initialised before submitting the address we need to manually call `autofill`

packages/client/src/components/Router.tsx
```diff
@@ -1,28 +1,27 @@
...
-    <BrowserRouter>
-      <ScrollToTop />
-      <Suspense fallback={<LoadingPage />}>
-        <Switch>
-          {redirectConfig.map((redirect) => (
-            <Redirect key={redirect.from} {...redirect} />
+const Router = () => (
+  // BrowserRouter has been moved to `../index.tsx` to make it accissible for other HOC's
+  <>
+    <ScrollToTop />
```
svenjens, line 11: Btw, do we still use the scroll to top?

svenjens (approved):

packages/client/src/components/Router.tsx
robinpiets: For sure. On every route. To reset the viewport.

packages/client/src/SessionContext.tsx
afjlambert: Not sure, looks great imo!

afjlambert (approved):

PR #728, opened 2020-12-08
Bugs: Incorrect loading state and incorrect alert show. 

svenjens:
Description 
This PR fixes a bug where permit alerts are overwritten for each permit. 
And fixes a bug where loading isn't shown. 

Please make sure:
- [x] to add a label


packages/client/public/imtr/transformed/2eYY7McaiRgfnERo9.json
```diff
@@ -1,7 +1,7 @@
 {
```
robinpiets, line 1: Can you remove all the new imtr files from this PR?

packages/client/src/pages/CheckerPage.js
```diff
@@ -162,7 +164,7 @@ const CheckerPage = ({ checker, matomoTrackEvent, resetChecker, topic }) => {
       newQuestionIndex = value;
       // Matomo event props
       action = actions.EDIT_QUESTION;
-      eventName = checker.stack[newQuestionIndex].text;
+      eventName = checker.stack[newQuestionIndex]?.text || t("common.unknown");
```
robinpiets: This is strange, IMHO, because on line 173 (`action = checker.stack[questionIndex].text`) it is not checked liked this.

Is the reason we're doing it like this because we get errors with this? In that case it's better to check if `checker.stack[newQuestionIndex]` exists in the first place. Because then we don't need events like this, right? Shall we discuss this tomorrow?
 
It would be better to not to call the whole event if the new index doesn't exist. Now we're calling a

packages/client/src/components/Questions.tsx
```diff
@@ -238,7 +238,10 @@ const Questions: React.FC<
             .getDecisiveInputs()
             .pop() as ImtrQuestion;
           const index = checker.stack.indexOf(decisiveQuestion);
-          permitsPerQuestion[index] = outcomeType;
+
+          if (!permitsPerQuestion[index]) {
+            permitsPerQuestion[index] = outcomeType;
+          }
```
robinpiets, line 243: 👍

robinpiets (changes requested):
Can you update the PR-title accordingly? And can you add only the checkers you need to this PR? Now it's very annoying to detect your changes. It's even better to `npm run imtr` on the base branch, commit and push and then merge the base branch into this PR. That way you don't check-in any imtr files.

robinpiets (changes requested):
⬆️

packages/client/src/pages/CheckerPage.js
svenjens: Let's send a error to sentry if newQuestionIndex is not found.

packages/client/src/pages/CheckerPage.js
```diff
@@ -160,13 +161,25 @@ const CheckerPage = ({ checker, matomoTrackEvent, resetChecker, topic }) => {
     if (Number.isInteger(value)) {
       // Edit specific question index (value), go directly to this new question index
       newQuestionIndex = value;
+      if (!checker.stack[newQuestionIndex]) {
+        captureException(
+          `Go to question, question with index: ${newQuestionIndex} not found`
+        );
+        return null;
```
robinpiets: It could just `return;` instead of `return null;`, because it's a plain function and not a component.

robinpiets (changes requested):
I found a bug with not showing the alert.

To reproduce:
- go to checker demolition
- address: 1055xd
- give answers: No, Yes

The result is NEED_PERMIT without any sign of an Alert

robinpiets (approved):
Nice fixes!

PR #730, opened 2020-12-09
Always print process.env debug information in footer

afjlambert:


afjlambert:
@robinpiets can you check this out. It's based on your branch.

robinpiets (approved):
Awesome!

PR #732, opened 2020-12-10
Replaced instances of Conclusion to Outcome

robinpiets:
Since we changed the text from "Conclusion" to "Outcome" we decided to change all corresponding Component files and instances as well.

Only these haven't change:
- the values for Matomo tracking events, because then we need to combine all "conclusion" with "outcome" and that's such a waste. So on Matomo it will still be "naar conclusie" for example, but in the code we use the key `GOTO_OUTCOME`
- the value for the section we use in Session Storage, because otherwise people might be stuck in the app when we release. The key has changed from `CONCLUSION` to `OUTCOME` though
- `__mocks__/with-conclusion-description.dmn.json` in parser.test.js on the `imtr` package (if you want these changed as well, let me know)
- some `conclusion` variables in `getOutcomesToDisplay()` in checker.ts on the `imtr-client` package (couldn't change because `outcome` is already defined. Anyone knows a better name, let me know)
- the `isConclusive()` function in checker.ts on the `imtr-client` package (we can change this to `hasOutcome()`, but I liked the other one better)


afjlambert (approved):

afjlambert:
@svenjens if you approve, please merge

svenjens (approved):

PR #741, opened 2020-12-17
Add db poc

afjlambert:
Description 

Please make sure:
- [ ] to add a label

### DIFF
--- a/packages/graphql/app.js
+++ b/packages/graphql/app.js
@@ -1,6 +1,7 @@
 const express = require("express");
 const morgan = require("morgan");
 const bodyParser = require("body-parser");
+const { Pool, Client } = require("pg");
 
 const config = require("./config");
 const { server } = require("./src/graphql");
@@ -11,6 +12,8 @@ const app = secure(express());
 if (process.env.NODE_ENV !== "production") {
   development(app);
 }
+
+// TODO: const pool = new Pool()
 
 // Basic access log, BUT no permission to log IP
 app.use(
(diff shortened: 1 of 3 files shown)


afjlambert:
@robinpiets Can you get this working locally? Let's discuss how to do the actual implementation.

packages/graphql/app.js
```diff
@@ -29,6 +32,25 @@ app.use((err, req, res, next) => {
   res.status(500).send("Something broke!");
 });
 
+const c = {
+  host: "db01.acc.dsq.amsterdam.nl",
+  database: "dsr_vergunningchecker_dev",
+  // port: 5432,
+  user: "vergunningchecker",
+  password: process.env.DB_PASS,
```
robinpiets, line 40: This should be a more specific variable in case we get multiple database passwords. eg: `DB_PASS_DSR`, `DB_PASS_VEGETATION`, etc (after POC is finished)

packages/graphql/app.js
```diff
@@ -29,6 +32,25 @@ app.use((err, req, res, next) => {
...
+};
+
+// db poc
+app.use(`/db`, async (_, resp) => {
+  const client = new Client(c);
+  await client.connect();
+  const res = await client.query("SELECT * FROM grn_vegetatieobject");
+  console.log(res.rows);
```
robinpiets, line 48: temp log (after POC is finished)

robinpiets (changes requested):
POC Works! I'm getting results from the db. Let's discuss implementation.

image

packages/graphql/app.js
afjlambert: What does DSR stand for?
Vegetation is maybe too narrow as we'll add more resources later.

Alternative could be `DSO_DB_PASS`?

PR #758, opened 2021-01-05
Section components in isolation and general refactor

robinpiets:
This is the refactor in CheckerPage that makes sure "Section Components don't not know about the current state of the CheckerPage". The CheckerPage used to render the three different sections (**Location**, **Questions**, **Outcome**) hardcoded and contained functions like `setActiveState("location")` and `setFinishedState("questions")`.

In this PR the CheckerPage is changed to dynamically construct the three sections based and uses functions like `goToNextSection()` and `completeSection()` to make sure underlying components don't know about the current state. This makes the CheckerPage way more robust and it's easier to add or remove sections in the future.

Done:
- all individual component children of CheckerPage don't know anything about the state of the CheckerPage
- checkers without data-needs skip rendering the Location Section
- checkers without any questions skip rendering the Question Section (💪🏽  - [see this check](http://localhost:3000/2eYY7McaiRgfnERo9/vragen-en-uitkomst))
- each section has their own Section file which contains the logic what to render
- fixed some bugs with autofill, the Debug table, color styling of `StepByStepNavigation`
- added translations to files
- updated function names in `Questions` and `Question`
- refactored Questions (marked all TODO's in [this card](https://trello.com/c/ZWvyG3Xi/209-refactor-questions-tests-wip))
- refactored boolean answer logic with a function called `getAnswerLabel()`

Please make sure:
- [x] to add a label


packages/client/src/types.ts
```diff
@@ -31,16 +31,47 @@ export type AddressType = {
...
  */
+
 export type TopicData = {
-  activeComponents?: string[];
   address: Address;
   answers: {
     [id: string]: Answer;
   };
-  finishedComponents: string[];
+  timesCheckerLoaded: number;
```
svenjens, line 69: there isn't a other way?

packages/client/src/utils/test-utils.tsx
```diff
@@ -16,7 +16,10 @@ import { SessionProvider } from "../SessionContext";
 
 export * from "@testing-library/react";
 
-dotenv.config();
+dotenv.config({
+  // With this option you can suppress all the console outputs except errors and deprecation warnings.
+  silent: true,
```
svenjens, line 21: What kind of outputs? Can you give a example

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,313 +1,336 @@
...
 
-    setTopicData({ activeComponents: [component] });
-  };
+      // Autocomplete the location section in case a new checker with data needs was started with a stored address
+      if (address && hasDataNeeds && !sectionData.length) {
+        // Complete the location section
+        defaultSections[0].isActive = false;
+        defaultSections[0].isCompleted = true;
+        // Activate the question section
+        defaultSections[1].isActive = true;
```
svenjens: Right now the situation is still specific for our use case. 

How does this work on default? 
When only specifying sections and start at 0

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,313 +1,336 @@
...
+          sectionFunctions,
+        });
 
-  // Callback to go to the Outcome section
-  // `false` is to prevent unexpected click, hover and focus states on already active section
-  const handleOutcomeClick =
-    !isActive(sections.OUTCOME) && isFinished(sections.QUESTIONS)
-      ? () => setActiveState(sections.OUTCOME)
-      : false;
+      const handleOnClick = () => {
```
svenjens: Can't we just call activeSection directly?

packages/client/src/debug/DebugDecisionTable.tsx
```diff
@@ -37,21 +38,53 @@ const DebugDecisionTable: FunctionComponent = () => {
...
-            <>
-              <h1>Autofill</h1>
-              {autofilled.map((q: Question) => (
-                <p key={q.id}>
-                  <QuestionSummary question={q} />: <Answer question={q} />
-                </p>
-              ))}
-            </>
-          )}
+    <HideForPrint>
```
robinpiets, line 43: This file is only ordered differently and has `HideForPrint` as wrapper, that's it

packages/client/src/utils/test-utils.tsx
```diff
@@ -16,7 +16,10 @@ import { SessionProvider } from "../SessionContext";
 
 export * from "@testing-library/react";
 
-dotenv.config();
+dotenv.config({
+  // With this option you can suppress all the console outputs except errors and deprecation warnings.
```
robinpiets, line 20: This gets rid a lot of warnings during `npm t`

packages/client/src/types.ts
robinpiets: Do you know a way to detect a first time loader?

packages/client/src/utils/test-utils.tsx
robinpiets: These kind of warnings: 
```
 PASS  src/components/Location/LocationSummary.test.tsx (8.079s)
  ● Console

    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_VERSION" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_GRAPHQL_API_URL" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_GIT_BRANCH" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_GIT_SHA" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_JOB_NAME" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_JENKINS_URL" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_BUILD_URL" is already defined in `process.env` and will not be overwritten
    console.warn node_modules/dotenv-flow/lib/dotenv-flow.js:74
      dotenv-flow: "REACT_APP_BUILD_NUMBER" is already defined in `process.env` and will not be overwritten
```

packages/client/src/pages/CheckerPage.tsx
robinpiets: Thanks. I've updated this

packages/client/src/utils/test-utils.tsx
robinpiets: image

robinpiets: Now it's this 🚀 

image

packages/client/src/pages/CheckerPage.tsx
robinpiets: Good point. Fixed this in [this commit](https://github.com/Amsterdam/vergunningcheck/pull/758/commits/b9fd958e28df87022ef2bcbcde2f8104b5a1bc29)

svenjens (approved):

packages/client/src/types.ts
robinpiets: ping @svenjens

svenjens: Not right now, but i almost can't imagine there wasn't something except changing the initial useEffect. 
But this way is pretty clear.

svenjens:
The 'vragen' header is supposed to be in the gray section. 
Screenshot 2021-01-20 at 14 45 20
Screenshot 2021-01-20 at 14 45 15

svenjens (changes requested):
Almost there :-) Please check my comments.

robinpiets:
> The 'vragen' header is supposed to be in the gray section.
> Screenshot 2021-01-20 at 14 45 20
> Screenshot 2021-01-20 at 14 45 15



> Almost there :-) Please check my comments.

These issues are fixed now.

svenjens (approved):

packages/client/src/components/Answers.tsx
```diff
@@ -1,62 +1,68 @@
...
-};
-
 type AnswersProps = {
-  answers?: AnswerProps[];
-  errors: any;
-  onChange: (e: FormEvent<HTMLInputElement>) => void;
-  questionId: string;
-  questionIndex: number;
-  userAnswer?: string;
+  errors: FieldErrors; // This prop needs to be passed down, because the useForm() hook failes fetching `errors` in this component
```
afjlambert: Small typo; "fails"

packages/client/src/components/Outcome/NewCheckerModal.tsx
```diff
@@ -58,33 +58,34 @@ const NewCheckerModal: FunctionComponent = () => {
...
-      // Set the the new session data for `doSaveAddress`
-      const topicSessionWithSavedAddress = {
+      // Set the new topic data with the correct `type` and potentially the `address`
+      const newTopicData = {
         ...defaultTopicSession,
-        activeComponents: [sections.QUESTIONS],
-        address: topicData.address,
-        finishedComponents: [sections.LOCATION_INPUT],
+        address: doSaveAddress ? topicData.address : null,
+        timesCheckerLoaded: checkerSlug === slug ? 2 : 0,
```
afjlambert: Not sure what's going on here.

packages/client/src/components/Location/LocationSection.tsx
```diff
@@ -0,0 +1,140 @@
...
+    setTopicData,
+    topicData: { address, sectionData, timesCheckerLoaded },
+  } = useTopicData();
+  const { matomoTrackEvent } = useTracking();
+  const { t } = useTranslation();
+
+  const {
+    currentSection: { isActive, isCompleted },
+    sectionFunctions: { goToNextSection },
+  } = props;
```
afjlambert, line 26: Nice!

packages/client/src/components/Outcome/OutcomeSection.tsx
```diff
@@ -0,0 +1,79 @@
...
+import { StepByStepItem } from "../StepByStepNavigation";
+import { OutcomeContent } from "./";
+
+const OutcomeWrapper = styled.div`
+  @media screen {
+    padding-bottom: ${themeSpacing(5)};
+  }
+`;
+
+const OutcomeSection: FunctionComponent<SectionComponent> = (props) => {
```
afjlambert, line 21: Like it!
Question though; what's the difference of React.FC vs FuctionComponent? If they are the same we should be consistent imo.

packages/client/src/components/StepByStepNavigation/StepByStepItem.tsx
```diff
@@ -87,10 +87,10 @@ const StepByStepItem: FunctionComponent<
         href,
       }}
       data-testid={STEPBYSTEPITEM}
-      aria-label={clickable ? heading : ""}
+      aria-label={clickable ? heading : ""} // @TODO: Refactor with `??`
```
afjlambert, line 90: clickable sounds more like a boolean, maybe undefined but not nullish imho

packages/client/src/debug/DebugDecisionTable.tsx
```diff
@@ -30,28 +31,62 @@ const DebugDecisionTable: FunctionComponent = () => {
   const { topicData } = useTopicData();
   const decisionId = "dummy";
 
-  if (!checker || !checker.permits) {
+  // Do not render without checker or when testing with Jest
+  if (!checker?.permits || process.env.JEST_WORKER_ID) {
     return <></>;
```
afjlambert: return null?

packages/client/src/i18n/nl.ts
```diff
@@ -2,16 +2,19 @@ export default {
   translation: {
     common: {
       "address loading": "Wij zoeken het adres.",
+      edit: "Wijzig",
       "housenumber label": "Huisnummer + toevoeging",
+      next: "Volgende",
+      no: "Nee",
+      "no valid postalcode":
```
afjlambert, line 9: I thought we did camelCase for keys in this file.

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,230 +1,220 @@
...
+
+const CheckerPage: FunctionComponent = () => {
   const { checker } = useChecker();
-  const { text } = useTopic();
-  const slug = useSlug();
-  const { matomoTrackEvent } = useTracking();
-  const sessionContext = useContext(SessionContext);
 
-  const hasDataNeeds = !!getDataNeed(checker);
+  const defaultSections: SectionObject[] = [
```
afjlambert: very cool

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,230 +1,220 @@
...
+    // Location Section:
+    {
+      ...defaultSectionData,
+      component: (props) => {
+        return <LocationSection {...props} />;
+      },
+    },
```
afjlambert, line 39: I'd expect Location-section to be conditional based on dataNeeds. Can you explain why this section is always added? Is `defaultSections` sometimes not used?

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,230 +1,220 @@
...
+  }, [checker]);
+
+  const getActiveSectionIndex = () => {
+    const activeSection = sectionsRef.current.filter(
+      (currentSection) => currentSection.isActive === true
+    );
+    if (!activeSection || activeSection.length !== 1) {
+      // Always expect one section to be active
+      const error = `"getActiveSectionIndex()": Expected one active section, got "${activeSection.length}"`;
+      captureException(error);
```
afjlambert, line 132: Should you not return here?

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,230 +1,220 @@
...
+    }
+    return activeSection[0].index;
```
afjlambert, line 137: this will throw an error if no section is active

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,230 +1,220 @@
...
-            (c: string) => !allComponents.includes(c)
-          ) || []
-        : [...finishedComponents, ...allComponents];
-
-    setTopicData({ finishedComponents: newFinishedComponents });
-  };
-
-  const isFinished = (component: string[] | string) =>
-    isActive(component, true);
+  const activateSection = (
```
afjlambert: Can you rename to something more related to what it does? Eg `addActiveSection`?

afjlambert: activateSectionIndex(index... ?

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -1,230 +1,220 @@
...
-        return;
-      }
-
-      // Matomo event props
-      action = checker.stack[questionIndex].text;
-      eventName =
-        value === "next"
-          ? eventNames.GOTO_NEXT_QUESTION
-          : eventNames.GOTO_PREV_QUESTION;
+  const completeSection = (
```
afjlambert, line 167: Shouln't this section stuff be part of a section-wrapper component of some kind.
What if we have a new page (besides CheckerPage for the one pager for example)? Then we'd also need a completeSection right?

packages/client/src/pages/CheckerPage.tsx
```diff
@@ -233,81 +223,31 @@ const CheckerPage = () => {
...
-            {hasDataNeeds && isActive(sections.LOCATION_INPUT) && (
-              <LocationInput
-                {...{
-                  handleNewAddressSubmit,
-                }}
-              />
-            )}
-            {hasDataNeeds && !isActive(sections.LOCATION_INPUT) && (
-              <LocationSummary showEditLocationModal />
+            {sectionsRef.current.map(
```
afjlambert, line 236: 🙌

packages/client/src/types.ts
```diff
@@ -31,16 +32,42 @@ export type AddressType = {
...
  * Context and session types
  */
 export type TopicData = {
-  activeComponents?: string[];
   address: Address;
   answers: {
     [id: string]: Answer;
   };
-  finishedComponents: string[];
+  timesCheckerLoaded: number;
```
afjlambert, line 69: Can't this be a boolean? checkerLoaded (true / false)?

packages/client/src/types.ts
```diff
@@ -91,8 +118,19 @@ export type Topic = OloTopic | IMTRTopic | RedirectToOloTopic;
 /**
  * Checker related types
  */
-export type BooleanOption = {
+export type AnswerOptions = {
```
afjlambert: Why plural? Isn't this one AnswerOption?

packages/client/src/utils/index.ts
```diff
@@ -104,3 +114,36 @@ export const isValidPostalcode = (value?: string) => {
...
+  },
+];
+
+/**
+ *
+ * This function can be used to get the labels of boolean questions
+ *
+ * @param {ClientSimpleType} answer
+ */
+export const getAnswerLabel = (answer: ClientSimpleType) => {
```
afjlambert, line 163: nice setup

packages/client/src/utils/index.ts
```diff
@@ -104,3 +114,36 @@ export const isValidPostalcode = (value?: string) => {
 export const removeQueryStrings = (value: string) => {
   return value.split("?")[0];
 };
+
+/**
+ *
+ * These are the hardcoded values and label for boolean questions
+ *
+ */
+export const booleanOptions: AnswerOptions[] = [
```
afjlambert: useful

packages/graphql/index.ts
```diff
@@ -0,0 +1,2 @@
+// This is an empty typescript file (which lives next to tsconfig.json) to satisfy my IDE.
```
afjlambert, line 1: Did you try this solution? 
https://stackoverflow.com/a/60234067

packages/imtr-client/src/models/checker.ts
```diff
@@ -78,6 +78,11 @@ export default class Checker {
       throw Error("Answers must be of type object");
     }
 
+    // Skip when answers are empty (to prevent error with checkers without questions to render)
```
afjlambert, line 81: Are checkers without questions even supported? :-/

afjlambert (changes requested):
Overall it looks good.
Main improvement is extracting all generic section-functions to a separate component. Then CheckerPage can do only the stuff that's non-generic.

packages/client/src/components/Outcome/OutcomeSection.tsx
robinpiets: `React.FC` stands for `FuctionComponent`. So it's the same. `React.FC` has been replaced throughout the whole app so this is consistent now.

packages/imtr-client/src/models/checker.ts
robinpiets: Yes, checkout this one: http://localhost:3000/2eYY7McaiRgfnERo9/vragen-en-uitkomst

robinpiets: So that means address leaves to rendering no questions because they are answered by autofill...

packages/client/src/components/Outcome/NewCheckerModal.tsx
robinpiets: I've added comments now

packages/client/src/components/StepByStepNavigation/StepByStepItem.tsx
robinpiets: Thanks, I've placed this comment in the right ToDo list on Trello. I will do this later and not on this PR

packages/client/src/debug/DebugDecisionTable.tsx
robinpiets: Yeah, better

packages/client/src/i18n/nl.ts
robinpiets: no we didn't, it's been setup to support either lowercase string or lowercase strings with spaces as keys

packages/client/src/pages/CheckerPage.tsx
robinpiets: Correct. At first i've implemented all the conditions inside the CheckerPage, but then I realised that the CheckerPage was getting way to much logic and complexity. So I've decided that each step should conditionally render it's own section content, based on its needs and based on whether or it's active or not. So Location Section is loaded by default and handles the question: "Is there data-needs and do I need to render?" and same for the Question Section, "Are there any questions to render?" That way the Section files handle all the logic and complexity.

If you want we can discuss offline

packages/client/src/pages/CheckerPage.tsx
robinpiets: True! Good point, this used to be a `throw new Error()`, but i've replaced this because it was annoying. A return will be added

packages/client/src/pages/CheckerPage.tsx
robinpiets: Thanks, good catch!

packages/client/src/pages/CheckerPage.tsx
robinpiets: changed to `changeActiveSection` as discussed offline

packages/client/src/pages/CheckerPage.tsx
robinpiets: as discussed, will do this on a new PR next sprint

packages/client/src/types.ts
robinpiets: No unfortunately I need to know if `timesCheckerLoaded === 0` (using the NewCheckerModal) `timesCheckerLoaded === 1` (on first load via normal route from Intro Page) or `timesCheckerLoaded > 1` (when the page reloaded). So a boolean doesn't do. But I'm very open to discuss other options because I don't really like this way either.

packages/client/src/types.ts
robinpiets: Because I was thinking: one answer have multiple options: formValue, label and value... I don't like this as well. The name AnswerOption(s) seem to be a combination of Answer (from frontend) and options (from IMTR).

Let's just make this `Answer`. What do you think?

robinpiets: Plain `Answer` makes a lot more sense and importing `Answer` from `imtr-client` will be imported like `{ Answer as IMTRAnswer }`

packages/graphql/index.ts
robinpiets: Yes I tried, but it doesn't matter, because in `src/` there aren't any TS files. I think the empty file is nice temporary solution until we convert to TS. Or we can delete the tsconfig file... 

Let's discuss later

packages/client/src/types.ts
robinpiets: I've added this situation to the dev-meeting topics

packages/client/src/types.ts
robinpiets: I've added this topic to next dev-meeting

robinpiets: Resolving this for now

PR #778, opened 2021-01-25
Checker `fire-safety`

robinpiets:
This is the PR for the new configured checker `fire-safety`.

It's currently in draft because we need to verify it with users first


robinpiets:
Closing this PR, because #781 is replacing this one

PR #793, opened 2021-02-01
Added validateDmnText function to handle the new dmn:text values in IMTR

robinpiets:
Since today, the API output has changed. Some XML fields `dmn:text` had changed its value from the expected `"-"` to the unexpected `"not(null)"`. All permit-check outcomes resulted in 'permit-free'.

I've added a function to the imtr parser to replace these values back to its expected value. I've also added some tests, although tests are not working in the imtr package, see `packages/imtr/package.json`

image


svenjens (approved):

robinpiets:
Closing this PR, because this the bug has been fixed at FL

PR #801, opened 2021-02-09
Fixed regex bug on Safari

svenjens:
This hot fix takes care of a regex bug in safari.


robinpiets (approved):
It works! Tested in Safari (MacOS / iOS), Chrome (different versions on Windows / Mac), Edge and Firefox

PR #812, opened 2021-02-17
Improved accessibility and readability to comply with WCAG

robinpiets:
I've added some quick fixes to improve accessibility and readability according to a WCAG quick scan

### DIFF
--- a/packages/client/src/components/Footer/Footer.tsx
+++ b/packages/client/src/components/Footer/Footer.tsx
@@ -29,6 +29,7 @@ const FirstColumn = () => (
         <PhoneNumber
           darkBackground
           eventName={sections.FOOTER}
+          underline
           variant={null}
         />
       </strong>{" "}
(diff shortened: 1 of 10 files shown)


svenjens (approved):
Nice improvements! @svenjens approves!

packages/client/src/components/Footer/Footer.tsx
```diff
@@ -29,6 +29,7 @@ const FirstColumn = () => (
         <PhoneNumber
           darkBackground
           eventName={sections.FOOTER}
+          underline
```
afjlambert, line 32: underline should not be a property of PhoneNumber imho. If you want to render the phone number in print for example, it does not make sense to underline. Underline is a property of a link, not a piece of text.

packages/client/src/components/HeaderStyles.ts
```diff
@@ -46,6 +46,7 @@ export const StyledLogo = styled.span`
   cursor: pointer;
   background-size: cover;
   background-image: url("${LogoMobile}");
+  text-indent: -9999px;
```
afjlambert: Is this related to wag? Maybe add a comment if so.

packages/client/src/components/Link.tsx
```diff
@@ -23,7 +15,7 @@ const Link = ({
   eventName,
   href,
   ...rest
-}: LinkProps) => {
+}: LinkProps & ComponentProps<typeof StyledComponentLink>) => {
```
afjlambert: Much improve 👍

packages/client/src/components/PhoneNumber.tsx
```diff
@@ -1,12 +1,20 @@
...
 import { PHONE_NUMBER } from "../utils/test-ids";
 import Link from "./Link";
 
-const Wrapper = styled.span`
+type WrapperProps = { underline?: boolean };
+
+const Wrapper = styled.span<WrapperProps>`
   white-space: nowrap;
+
+  ${({ underline }) =>
```
afjlambert: Me like syntax 🙌

afjlambert (approved):
Small stuff, maybe change the underline on PhoneNumber?

packages/client/src/components/Footer/Footer.tsx
robinpiets: I understand what you mean, I will place the underline inside the Link. The PhoneNumber, however, still needs an underline prop

PR #813, opened 2021-02-17
Prevent a bug when refreshing a checker without getDataNeed

robinpiets:
PR #805 is now causing a bug for checkers without data-need [Brandveilig gebruik]: when refreshing without answering the first question, the question section is empty. This PR fixes that bug.

**Reproduce bug:**
1. Go to this [check](https://develop.chappie2.com/brandveilig-gebruik/vragen-en-uitkomst) on the develop branch
2. Refresh the page (without answering the first question)
3. The question section is completely empty

image

### DIFF
--- a/packages/client/src/components/Question/QuestionSection.tsx
+++ b/packages/client/src/components/Question/QuestionSection.tsx
@@ -1,6 +1,7 @@
 import React, { FunctionComponent, useEffect } from "react";
 import { useTranslation } from "react-i18next";
 
+import { getDataNeed } from "../../config/autofill";
 import { actions, sections } from "../../config/matomo";
 import { useChecker, useSlug, useTopicData, useTracking } from "../../hooks";
 import { SectionComponent } from "../../types";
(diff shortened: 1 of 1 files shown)


packages/client/src/components/Question/QuestionSection.tsx
```diff
@@ -80,6 +81,16 @@ const QuestionSection: FunctionComponent<SectionComponent> = (props) => {
     return null;
   }
 
+  // Prevent a bug when refreshing a checker without getDataNeed (only happens when first question remains unanswered)
+  if (
+    !getDataNeed(checker) &&
+    isActive &&
+    !isCompleted &&
+    checker.stack.length === 0
```
svenjens: can't we change this to a explaining const. 
```
const noDataNeedAndEmptyStack = !getDataNeed(checker) && isActive && !isCompleted && checker.stack.length === 0

if (noDataNeedAndEmptyStack) 
```

It's a bit hard to read right now.

svenjens (approved):

packages/client/src/components/Question/QuestionSection.tsx
robinpiets: I've implemented this now.

Just a side note: I'm not a big fan of variables only being used on one location. Also the name will never do justice, because in reality it should be `noDataNeedAndEmptyStackAndActiveButNotCompletedQuestionSection` (-:

robinpiets: @afjlambert, opinion?

afjlambert: I see your point. It is however the same as the regexes earlier. Sometimes naming things makes it very descriptive / easy to recognize. A good short comment also helps of course, downside is the comment is sometimes not update with the code 😊.
Small thing related to your comment, it is no longer a bug 😊 
PR looks good 👍

afjlambert (approved):

PR #814, opened 2021-02-17
Added FileUpload demo

robinpiets:
This PR is used for a UX test only. It adds a demo version of the FileUpload feature.

image


robinpiets:
Not needed anymore

PR #823, opened 2021-02-24
Husky configuration update for husky 5

svenjens:
Description 
This pr changes the husky configuration so its working with husky 5 precommit. 

Please make sure:
- [x] to add a label


.husky/pre-commit
```diff
@@ -0,0 +1 @@
+npx --no-install lerna run --concurrency 1 --stream precommit
```
afjlambert, line 1: Just curious; why did you add `npx --no-install ...`?

afjlambert (changes requested):
For me the hooks did not work after a `npm run cleaninstall`. I had to `npx husky install` first. We could add that to the docs, but maybe better is following the docs. It states we can add this to our `package.json`:
```
"scripts": {
    "postinstall": "husky install"
  }
```
See https://typicode.github.io/husky/#/

.husky/_/husky.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/sh
```
afjlambert: I think you can remove this husky.sh file. Why did you add this?
Reading the docs, I think it should be installed automatically in `.git/hooks`-folder after you `npx husky install`.

afjlambert (changes requested):

.husky/_/husky.sh
svenjens: This is the file that is added by installing husky. 

Husky created it. :-)

.husky/pre-commit
svenjens: Its running Lerna without installing in.
If we don't add npx --no-install it can't find lerna.

.husky/_/husky.sh
afjlambert: Is this for Windows users maybe? Seems to run fine without.
I think I installed husky in a clean project and it did not add this file 🤔

svenjens:
Let me check if I delete this file it still works

Op ma 1 mrt. 2021 om 12:34 schreef André Lambert <notifications@github.com>

robinpiets:
I'm getting massive change requests after running `npm i` in the root. 

image

robinpiets (approved):
👍🏽

svenjens:
> 👍🏽

@robinpiets  I see you added the husky .gitignore. I didn't get that file on install..

PR #846, opened 2021-03-16
Cleaned up files and updated tests

robinpiets:
Cleaned up files and updated tests


packages/client/src/routes.test.ts
```diff
@@ -1,17 +1,13 @@
 import slugify from "slugify";
 
-import { getslug, geturl } from "./routes";
+import { geturl } from "./routes";
 
 const text =
   "U gaat het dakraam plaatsen aan de achterkant van het gebouw. Ligt die kant aan een straat, fietspad of voetpad waar iedereen mag komen?";
 const slug =
```
svenjens, line 7: We don't use question slugs anymore right? Where is this slug being used.

robinpiets: This slug is just a test to see if the conversion works correctly in `slugify`

svenjens (approved):

PR #882, opened 2021-04-15
started implementation of provider

svenjens:
Description 

Please make sure:
- [ ] to add a label


afjlambert:
I'm not sure where you want to go with this 🤔 
If we model the graphql server the same way we modelled our json then we don't need to define any new types in the `client` package.
Let's discuss our approach for the setup.

